PR #10, opened 2015-04-30
Messages API

mandric:
WIP, opening for discussion.  Issue: https://github.com/medic/medic-webapp/issues/817


mandric:
Update handler call to `kutils.setTaskState` might not change anything, if we for example set to the same state.  Should we modify to check return values on that function so it doesn't set a state that is currently set?

PR #15, opened 2015-05-19
ODKforms support

mandric:
WIP: medic/medic-webapp/issues/886

Serve forms to ODK Collect

todo: tests, PUT and POST support


mandric:
Porting these changes to testing and then will merge from there.

PR #18, opened 2015-05-20
Pull form fields into a namespaced 'fields' property.

garethbowen:
Issue: medic/medic-webapp#957


tests/migrations/namespace-form-fields.js
```diff
@@ -0,0 +1,147 @@
...
+  var getConfig = sinon.stub(config, 'get').returns(forms);
+  var insert = sinon.stub(db.medic, 'insert').callsArgWith(1, null, null);
+  migration.run(function(err) {
+    test.equals(err, undefined);
+    test.equals(getView.callCount, 1);
+    test.equals(getDoc.callCount, 1);
+    test.equals(getConfig.callCount, 1);
+    test.equals(insert.callCount, 1);
+    test.same(insert.args[0][0].fields, expected);
+    test.equals(insert.args[0][0].last_menstrual_period, null);
```
mandric: should these fields be null or undefined now?

garethbowen: `equals` tests with `==` so this assertion passes with either `null` or `undefined`. I could use `same` to be explicit about it if you'd prefer?

mandric: Well there is a difference, null will appear on the doc but undefined won't.

garethbowen: Right. They'll be undefined now as I `delete` them off the doc.

mandric: If we don't want them on the doc then we should probably assert they are undefined?

PR #19, opened 2015-06-29
Add bash script to check for bad translations of substitutions in {}

alxndrsn:
This will report if there are translations made of variables names in the
`translations/` files.  It should also break the build on Travis.

I'm not sure if the travis config will work, but hopefully Travis will answer that one.


alxndrsn:
@abbyad this is working as expected - it's failing because there are translations of `{{variables}}`

garethbowen:
@alxndrsn Nice one! I feel like it'd be a little nicer in a unit test than a bash script. Whaddya reckon?

browndav:
IMHO: I'd probably avoid `$(ls .)` since aliases and `$LS_COLORS` can pretty easily hose things up. Maybe just `for tran in *.properties`?

alxndrsn:
@browndav thanks, will fix that
@garethbowen i know what you mean, but i have no idea the priority of this and doing it in shell is minutes work, whereas doing in JS would take me a bit longer

mandric:
Can we fix the translations and merge this?

mandric:
Also any chance we can get 4 or 2 space indentation?

garethbowen:
@alxndrsn It'd be nice to get this merged. Could you have a look at fixing the translations and changing the indentation (as Milan pointed out). Also it might be better to move the script into `/tests` to keep it away from the production code?

garethbowen:
@alxndrsn Did you follow [the instructions](https://github.com/medic/medic-docs/blob/master/md/dev/translations.md) when updating the translations? If not the bad translations will come back...

PR #24, opened 2015-07-23
1098 escape html special chars in redirect URLs

alxndrsn:


garethbowen:
Looks good. Make it so.

PR #25, opened 2015-07-23
1098 check redirect URL on server-side before templating login page

alxndrsn:


garethbowen:
This looks good to me. It will break the existing code that uses absolute paths, including
- https://github.com/medic/medic-api/blob/develop/server.js#L462
- https://github.com/medic/medic-api/blob/develop/server.js#L539
- https://github.com/medic/medic-webapp/blob/develop/static/js/modules/manage-session.js#L9

garethbowen:
It might be easier to strip the host off in API with node URL parsing than doing it in kanso, angular, and node. Maybe when rendering the template just pass in the URL without the host, port, etc and then the rest of the code can call with absolute or relative paths.

browndav:
There's https://github.com/unshiftio/url-parse that might be a viable for hostname removal or other types of super-strict checking.

garethbowen:
Or https://nodejs.org/api/url.html for server side.

garethbowen:
I'd like to see some unit testing for this. If you pull the login stuff, or at least the URL parsing, into a module it'll be easy enough to test.

PR #30, opened 2015-08-17
Fix part2.1

alxndrsn:
Fix for #25


browndav:
Other than the above comment, I did a quick read-through and this LGTM.

alxndrsn:
Good catch.

PR #38, opened 2016-02-04
sanitize requests in the forms controller. 

mandric:
Added a `db.sanitizeResponse` that can be called from any controller.  The only
problematic function right now is `forms.getForm` because it returned headers
coming from nano.  Sanitized those headers by removing `uri` and `statusCode`
and wrote tests around it.  This could potentially leak auth information to the
client.  See https://github.com/dscape/nano/blob/master/lib/nano.js#L195

TODO: open issue in nano project and patch

Issue medic/medic-webapp#1844


mandric:
Doh, meant to say "sanitize responses" in the commit title.

Issue #36, opened 2016-02-02
Support extracting and update secondary ddocs


SCdF:
We want to be able to have multiple ddocs, but still use one master ddoc. We're going to solve this by having extra ddocs as attachments to the main one, and write some code in medic-api that on boot or ddoc change, evaluate those ddocs to see if they need updating.

SCdF:
https://github.com/medic/medic-api/pull/39

garethbowen:
Nice work. I've added a couple of comments including one bug (I think?). I think it's a bit verbose too - maybe remove the logging for ddocs that aren't new/modified?

SCdF:
Hey @garethbowen I've made some changes based on your suggestions, let's talk about it tonight

garethbowen:
LGTM. @SCdF merge when ready!

SCdF:
Merged

PR #39, opened 2016-02-05
Automatic DDoc extraction #36

SCdF:
This feature supports our ability to have muiltiple DDocs being served out of
the same Kanso project.

On the `medic-webapp` side, all extra Ddocs should be JSON files (with their
correct `_id` etc) inside the `ddocs/compiled` directory. This is then picked
up by Kanso as an attachment and pushed to Couch.

When `medic-api` runs it will run this code. This code looks through all
attached Ddocs, and re-uploads them if it thinks it should.

### DIFF
--- a/ddoc-extraction.js
+++ b/ddoc-extraction.js
@@ -16,8 +16,6 @@ var extractDdocs = function(ddoc, callback) {
     return name.indexOf('ddocs/compiled/') === 0;
   });
 
-  console.log('Found ' + attachmentNames.length + ' attached ddocs');
-
   callback(null, ddoc._rev, attachmentNames);
 };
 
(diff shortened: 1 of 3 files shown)


SCdF:
So I'm not a massive fan of how this turned out FWIW, but I think it's worth taking a look at. The more I dug into attachments the more I realised that:
 - This code doesn't need to be specific to ddocs
 - This code doesn't really need to check and store the parent rev, and instead could probably just use the attachment hash, either by storing it or calculating it

SCdF:
Build failures appear to be based on node version, especially since the two different builds fail (for two different node versions) fail in different ways.

@garethbowen / @browndav what is the official node / npm / whatever version(s) that `medic-api` should run / pass on?

alxndrsn:
@SCdF should pass on versions configured on travis - currently 0.10, 0.12.

SCdF:
Yay!

SCdF:
Hey @garethbowen can you take a gander at this?

ddoc-extraction.js
```diff
@@ -0,0 +1,91 @@
...
+          }
+        }
+      });
+    }
+  });
+};
+
+module.exports = {
+  run: function(runComplete) {
+    console.log('Extracting ddocs…');
```
garethbowen: Ellipsis? Really?? :P

ddoc-extraction.js
```diff
@@ -0,0 +1,91 @@
...
+            _.partial(updateIfRequired, masterRevision),
+            callback);
+        }
+      ], function(err) {
+        if (err) {
+          console.log('Something went wrong trying to extract ddocs', err, console.trace());
+        } else {
+          console.log('…Done');
+        }
+        runComplete(err);
```
garethbowen: What does this do?

ddoc-extraction.js
```diff
@@ -0,0 +1,91 @@
...
+    return name.indexOf('ddocs/compiled/') === 0;
+  });
+
+  console.log('Found ' + attachmentNames.length + ' attached ddocs');
+
+  callback(null, ddoc._rev, attachmentNames);
+};
+
+var ddocNameFromAttachmentName = function(attachmentName) {
+  return '_design/' + /ddocs\/compiled\/(.+)\.json/.exec(attachmentName)[1];
```
garethbowen: I found this hard to read. Maybe pull out the RE as a named "constant"?

ddoc-extraction.js
```diff
@@ -0,0 +1,91 @@
...
+  console.log('Looking at', attachmentName);
+
+  db.medic.get(ddocName, function(ddocErr, ddoc) {
+    if (ddocErr && ddocErr.error !== 'not_found') {
+      callback(ddocErr);
+    } else {
+      db.medic.get('_design/medic/'+attachmentName, function(attachErr, attachedDdoc) {
+        if (attachErr) {
+          callback(attachErr);
+        } {
```
garethbowen: Missing `else`?

I like the style...

```
if (err) {
   return callback(err);
}
// non-error code
```
This keeps the code less indented, and short circuits on err immediately.

ddoc-extraction.js
SCdF: >_>

ddoc-extraction.js
SCdF: Honestly at this point it's a hook for the tests, since it gives them a way of knowing that the task is complete. In general though, it's just a callback hook so that if you care externally about knowing that the task is complete, you can.

ddoc-extraction.js
```diff
@@ -0,0 +1,92 @@
...
+    return name.indexOf('ddocs/compiled/') === 0;
+  });
+
+  console.log('Found ' + attachmentNames.length + ' attached ddocs');
+
+  callback(null, ddoc._rev, attachmentNames);
+};
+
+var ddocNameFromAttachmentName = function(attachmentName) {
+  var designDocNameFromFilePath = /ddocs\/compiled\/(.+)\.json/;
```
SCdF: Hey @garethbowen does that help? I didn't split it out originally because I was hoping that the function name would be clear enough that you didn't need to dig into the specifics. Let me know if that name makes no sense to you etc.

ddoc-extraction.js
SCdF: And yet the tests pass… clearly I didn't write enough

ddoc-extraction.js
garethbowen: Yeah I find that much easier. There were too many slashes on one line...

PR #44, opened 2016-02-11
Seems to make some things more consistent

alxndrsn:
@scdf i can't start api on `develop`; these changes help


SCdF:
I solved it with your other smaller suggestion instead, see: https://github.com/medic/medic-api/commit/c9d010ff9e1a975693d4a74cfcb4a3be764b06d4

PR #48, opened 2016-03-08
1764 create users

mandric:
Issue https://github.com/medic/medic-webapp/issues/1764


mandric:
If you're curious about creating users have a look here at the [Users section in API_v1.md](https://github.com/medic/medic-api/blob/1764-create-users/API_v1.md#users).  It's a bit rough so feedback appreciated.

PR #49, opened 2016-03-14
users api refactor

mandric:
Issue medic/medic-webapp#2046


controllers/users.js
```diff
@@ -0,0 +1,410 @@
...
+  ].indexOf(place.type) !== -1;
+};
+
+var getPlace = function(id, callback) {
+  db.medic.get(id, function(err, place) {
+    if (err) {
+      return callback(err);
+    }
+    if (!isAPlace(place)) {
+      return callback('Wrong type, this is not a place.');
```
garethbowen: Always return Errors so we get the stacktrace.

controllers/users.js
```diff
@@ -0,0 +1,410 @@
...
+  });
+};
+
+var createUser = function(data, response, callback) {
+  response = response || {};
+  var id = createID(data.username),
+      user = getUserUpdates(id, data);
+  db._users.insert(user, id, function(err, body) {
+    if (err) {
+      return callback('Failed to create user. ' + err);
```
garethbowen: This doesn't work well if `err` is not a string. Maybe just `callback(err)` or use multiple args `callback('Failed to create user.', err)`.

controllers/users.js
```diff
@@ -0,0 +1,410 @@
...
+      if (err) {
+        return callback(new Error('Failed to find place. ' + err));
+      }
+      // validate contact parent exists
+      self._getContactParent(data.contact.parent, function(err, facility) {
+        if (err) {
+          return callback(new Error('Failed to find contact parent. ' + err));
+        }
+        if (!self._hasParent(facility, data.place)) {
+          return callback(new Error('Contact is not within place.'));
```
garethbowen: Nice one!

controllers/users.js
```diff
@@ -0,0 +1,410 @@
...
+ * Warning: not doing validation of the contact data against a form yet.  The
+ * form is user defined in settings so being liberal with what gets saved to
+ * the database. Ideally CouchDB could validate a given object against a form
+ * in validate_doc_update.
+ */
+var createContact = function(data, response, callback) {
+  response = response || {};
+  db.medic.insert(data.contact, function(err, body) {
+    if (err) {
+      return callback('Failed to create contact. ' + err);
```
garethbowen: This won't work nicely if err is an Error not a string. Maybe just `callback(err)` instead?

controllers/users.js
```diff
@@ -0,0 +1,410 @@
...
+    callback(null, data, response);
+  });
+};
+
+var hasParent = function(facility, id) {
+  if (facility._id === id) {
+    return true;
+  }
+  // clone facility before modifying it
+  var f = JSON.parse(JSON.stringify(facility));
```
browndav: I think we can probably avoid writing to `facility` here, which would remove the need for the clone AFAICT.

mandric:
Added fixes for above comments.

Still couple nagging concerns:

So I should use the `facilities` view to fetch places rather than the simple `db.medic.get`?  Similarly with  getContactParent?

The only way ATM to create more couchdb admins is with the admin account in futon.

Redundant properties on user and user-settings docs:

`name`
`known`
`facility_id`

We have a rolesMap hard coded in the controller code. Do we have a place documented 
with what those roles actually do?

I have not modified `validate_doc_update` in medic-webapp yet.  Should I and any particular things you want to see validated there?

Which one should we use for rendering GET /api/v1/users?  Using a rule of thumb, if the property is available on the user's doc we use it.  Should we show an error if they are not the same? 

Currently there is no way to query for a single user detail. i.e. GET /api/v1/users/{{username}}

garethbowen:
> So I should use the facilities view to fetch places rather than the simple db.medic.get? Similarly with getContactParent?

I think either is fine. Whatever is most convenient for you I think. Performance isn't much of a concern here. We know the view will exist so it comes down to code quality.

> We have a rolesMap hard coded in the controller code. Do we have a place documented
with what those roles actually do? Should we include that in the API docs?

TBH I don't understand the roles map - I think it just makes the whole thing more confusing than it needs to be. Instead of `user.type -> array of roles -> array of permissions` I think we should have `user.type -> array of permissions`. This is approximately how it's implemented anyway. Anyway, this rant it outside the scope of this issue. It's fine for now.

> I have not modified validate_doc_update in medic-webapp yet. Should I and any particular things you want to see validated there?

Yes that's a good idea. Most importantly we should validate the perms of the person doing the update.

> Which one should we use for rendering GET /api/v1/users? Using a rule of thumb, if the property is available on the user's doc we use it. Should we show an error if they are not the same?

I guess this is talking about the redundant properties? Always use the one on _users if you have the choice. Don't worry about validating it here - if someone messes with their user settings locally it'll error when they try to replicate anything.

> Currently there is no way to query for a single user detail. i.e. GET /api/v1/users/{{username}}

That's ok for now.

mandric:
Also there are some new privileges I'm using here.  I should probably look at adding some defaults to the webapp?  Otherwise it's admins-only.

garethbowen:
Yes definitely add them to the webapp settings defaults even if they default to none. That means they will show up in the Configuration > Permissions admin screen. Otherwise you just have to know they're there.

mandric:
Are those permissions defaults still over here? https://github.com/medic/medic-webapp/blob/develop/packages/kujua-sms/views/lib/app_settings.js  :cringe:

garethbowen:
Yeah that's the one  :(

controllers/users.js
```diff
@@ -0,0 +1,361 @@
...
+    callback(null, data, response);
+  });
+};
+
+var hasParent = function(facility, id) {
+  // do not modify facility
+  var p = facility.parent;
+  if (facility._id === id) {
+    return true;
+  }
```
garethbowen: I don't think this `if` block is necessary - if you don't have it the first check in the `while` loop returns `true` anyway.

controllers/users.js
```diff
@@ -0,0 +1,361 @@
...
+        if (err) {
+          return cb(err);
+        }
+        user._deleted = true;
+        db.medic.insert(user, cb);
+      });
+    }
+  ], function(err) {
+    callback(err);
+  });
```
garethbowen: I'm not sure this should be parallel. Maybe delete the _users entry first and only if that succeeds move on to deleting the user-settings?

controllers/users.js
```diff
@@ -0,0 +1,361 @@
...
+   * @param {Object} data - request body
+   * @param {Function} callback
+   * @api public
+   */
+  createUser: function(data, callback) {
+    var self = this,
+        required = ['username', 'password', 'place', 'contact'],
+        missing = [],
+        response = {};
+    for (var i in required) {
```
garethbowen: Use `required.forEach(function(...) {});`

mandric:
@garethbowen currently here's the permissions checks that seems to apply to modifying user docs.  Remember this only applies to the medic db so we're only talking about user-settings docs, assuming we don't want to get into permissions for contacts (even though I'm creating those too).

in [validate_doc_update.js](https://github.com/medic/medic-webapp/blob/develop/lib/validate_doc_update.js#L24):

```
    // admins can do anything
    if (isDbAdmin(userCtx, secObj)
        || utils.isUserAdmin(userCtx)
        || utils.isUserDistrictAdmin(userCtx)) {
        return {
            allowed: true
        };
    }

    // only admins can delete
    if (oldDoc && newDoc._deleted) {
        return {
            allowed: false,
            message: 'You must be an admin to delete docs'
        };
    }

    // gateway and data_entry users can update
    if (utils.hasRole(userCtx, 'kujua_gateway')
        || utils.hasRole(userCtx, 'data_entry')) {
        return {
            allowed: true
        };
    }

```

in [kujua-utils.js](https://github.com/medic/medic-webapp/blob/develop/packages/kujua-utils/kujua-utils.js#L43):

```

exports.isUserAdmin = function(userCtx) {
    return exports.isUserNationalAdmin(userCtx) || exports.isDbAdmin(userCtx);
};

exports.isUserNationalAdmin = function(userCtx) {
    return exports.hasRole(userCtx, 'national_admin');
};

exports.isUserDistrictAdmin = function(userCtx) {
    return exports.hasRole(userCtx, 'district_admin');
};
```

So this means that as-is if you have one of the basic roles (national_admin, district_admin, kujua_gateway or data_entry) you can modify user-settings docs, but only admins can delete.  Do we need more granular restrictions?

Here's the default perms I added for for can_*_users:

```
    {
      "name": "can_view_users",
      "roles": [
        "national_admin",
        "district_admin"
      ]
    },
    {
      "name": "can_create_users",
      "roles": [
        "national_admin",
        "district_admin"
      ]
    },
    {
      "name": "can_update_users",
      "roles": [
        "national_admin",
        "district_admin"
      ]
    },
    {
      "name": "can_delete_users",
      "roles": [
        "national_admin",
        "district_admin"
      ]
    }
```

I think ideally we want validate_doc_update to check that the user has the permission before they can modify a user-settings doc? 

Maybe something like below.  Unfortunately this is not supported in validate_doc_update, because calling utils.hasPerm doesn't check the ddoc/current settings, it just has a hard code set of permissions that it checks.  This likely deserves another issue to refactor kujua-utils.hasPerm.

```
    // check user permissions to modify user-settings doc
    if (newDoc.type === 'user-settings') {
      if (newDoc._deleted) {
        if (utils.hasPerm(userCtx, 'can_delete_users')) {
          return {
            allowed: true
          }
        }
      } else if (oldDoc) { 
        if (utils.hasPerm(userCtx, 'can_update_users')) {
          return {
            allowed: true
          }
        }
      } else if (utils.hasPerm(userCtx, 'can_create_users')) {
        return {
          allowed: true
        }
      }
      return { 
        allowed: false,
        message: "Not authorized to modify this document."
      } 
    }

```

controllers/users.js
mandric: Yeah I was on the fence about that one. You can get stuck in a place unable to fully delete a user and making a username unusable.  I would prefer two calls to delete that might fix the problem rather than making it impossible to delete a user-settings doc because the doc in _users doesn't exist.

controllers/users.js
mandric: In the while loop we are only checking the parent not the contact itself.

controllers/users.js
```diff
@@ -0,0 +1,361 @@
...
+  // do not modify facility
+  var p = facility.parent;
+  if (facility._id === id) {
+    return true;
+  }
+  while (p) {
+    if (p._id === id) {
+      return true;
+    }
+      p = p.parent;
```
browndav: Whitespace issue.

controllers/users.js
browndav: I think the suggestion here is to do `var p = facility;`, remove the duplicate `id` test, and let the loop body sort it out since `p` would obviously be truthy upon entry into the `while`.

mandric: :beers:

PR #50, opened 2016-03-15
Add translation for DDoc version on About page

alxndrsn:
Issue: medic/medic-webapp#2084

### DIFF
--- a/translations/messages-en.properties
+++ b/translations/messages-en.properties
@@ -577,6 +577,7 @@ debug.db_info.title = Database Info
 debug.mode.title = Enable debug mode
 debug.mode.description = Debug mode will print information into the browser console to help developers diagnose problems with the application.  After changing this setting reload the app so it applies.
 android_app.version.title = Android App Version
+ddoc.version = DDoc Version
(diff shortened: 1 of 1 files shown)


translations/messages-en.properties
```diff
@@ -577,6 +577,7 @@ debug.db_info.title = Database Info
 debug.mode.title = Enable debug mode
 debug.mode.description = Debug mode will print information into the browser console to help developers diagnose problems with the application.  After changing this setting reload the app so it applies.
 android_app.version.title = Android App Version
+ddoc.version = DDoc Version
```
garethbowen: Maybe change the value to something a little more user friendly. What about "Application Version"?

abbyad: "Application Version" may be confounded with the "webapp version", or the "android app version". Does something like "Config version" work?

garethbowen: It's both the "webapp version" and the "config version" all in one magic number. I'm not sure how to express that in English...

alxndrsn: Possibly safer to stick with ddoc, then?

PR #51, opened 2016-03-16
added update support to users controller. 

mandric:
WIP medic/medic-webapp#2047

Do we want to enforce password-strength standards?

Couple weird errors to end user below... not sure if this is acceptable.

Place not found error:
```
curl -X POST -H 'content-type:application/json' -d '{"place":"blah"}' 'http://admin:secret@localhost:5988/api/v1/users/henry'
Server error: missing
```

When type is not in the list of supported types (see User Types list in doc). Type not found error:
```
curl -X POST -H 'content-type:application/json' -d '{"type":"blah"}' 'http://admin:secret@localhost:5988/api/v1/users/henry'
Server error: doc.roles can only contain strings
```


controllers/users.js
```diff
@@ -354,5 +396,65 @@ module.exports = {
...
+              }
+              user.facility_id = placeID;
+              settings.facility_id = placeID;
+              // update settings
+              self._updateUserSettings(userID, settings, cb);
+            });
+          });
+        // update request only included place but it is the same
+        } else if (_.isUndefined(data.type) && _.isUndefined(data.password)) {
+          return callback(new Error(
```
mandric: On the fence about what to do if the request is to only update a place but the new data is the same as existing data.

garethbowen: I don't think we should error here. If you have a script with 1000 users, some of which are up to date, you want to be able to just update all even if some stay the same.

mandric: Ok just do nothing as if the update happened (return 200)?

mandric: Actually there is also a potential weird case here if user.facility_id and userSettings.facility_id are not equal... I need to check that too.

server.js
```diff
@@ -424,6 +424,20 @@ app.post('/api/v1/users', jsonParser, function(req, res) {
...
+app.post('/api/v1/users/:username', jsonParser, function(req, res) {
+  auth.check(req, 'can_update_users', null, function(err) {
+    if (err) {
+      return serverUtils.error(err, req, res);
+    }
+    users.updateUser(req.params.username, req.body, function(err) {
+      if (err) {
+        return serverUtils.serverError(err, req, res);
+      }
+      res.json();
```
garethbowen: It'd be good to return something here... maybe just `res.json({ success: true });` ?

server.js
```diff
@@ -424,6 +424,20 @@ app.post('/api/v1/users', jsonParser, function(req, res) {
...
 });
 
+app.post('/api/v1/users/:username', jsonParser, function(req, res) {
+  auth.check(req, 'can_update_users', null, function(err) {
+    if (err) {
+      return serverUtils.error(err, req, res);
+    }
+    users.updateUser(req.params.username, req.body, function(err) {
+      if (err) {
+        return serverUtils.serverError(err, req, res);
```
garethbowen: I think serverError is a bit aggressive - usually this is reserved for unexpected exceptions (like bugs in the code, out of memory, etc). For example, we don't need to output to console.error just because someone supplied invalid args. Try and work out what the right http status code is and return a useful JSON response with `success: false` or something. Perhaps where the error is generated you could return a JS object instead, for example instead of...

```
return callback(new Error(
  'One of the following fields are required: type, password or place.'
));
```

You could use...

```
return callback({
  code: 400,
  message: 'One of the following fields are required: type, password or place.'
});
```

There is code in `serverUtils.error` to handle this but it would need updating if you choose to use it.

mandric: Nice suggestion, I knew I needed to rework some error handling and this is probably the way to go.

mandric:
Little better now:

```
curl -vX POST -H 'content-type:application/json' -d '{"place":"blah"}' 'http://admin:secret@localhost:5988/api/v1/users/henry'

HTTP/1.1 400 Bad Request
Failed to find place.
```

controllers/users.js
mandric: I modified this to just always set (even set the same value) if place passes validation.

mandric:
Ok getting closer on this PR.  We can update any field now.  Questions:

No validation on some fields like `known` etc, you can set known to 3 for example.  Would be nice to refactor and have a way to define validation on every field and be DRY.

200 http response ok? Do we want a JSON response similar to create.

Do we save roles on both docs (user and user-settings)?

mandric:
Maybe add a `validate_doc_update` that always checks docs for a `type` property. Then each type should have some basic validation being done, like check the `known` field is boolean.

garethbowen:
> Do we save roles on both docs (user and user-settings)? 

Yes we have to. Save on `user` because `user-settings` can't be trusted as it can be modified client side. Save on `user-settings` because `user` can't be replicated and we need to be able to read it client side.

> 200 http response ok? Do we want a JSON response similar to create?

A simple JSON response is nice IMO.

mandric:
Any thoughts about updating `validate_doc_update` here?

package.json
```diff
@@ -42,6 +43,7 @@
   },
   "scripts": {
     "start": "node server.js",
-    "test": "grunt test"
+    "test": "grunt test",
+    "updatetoc": "./node_modules/.bin/doctoc --maxlevel 2 API_v1.md"
```
garethbowen: Nice one!

PR #55, opened 2016-04-06
support places better

mandric:
Issue: medic/medic-webapp#2046

### DIFF
--- a/controllers/places.js
+++ b/controllers/places.js
@@ -1,5 +1,4 @@
-var async = require('async'),
-    _ = require('underscore'),
+var _ = require('underscore'),
     db = require('../db');
 
 var getPlace = function(id, callback) {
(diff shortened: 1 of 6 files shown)


controllers/places.js
```diff
@@ -0,0 +1,111 @@
...
+  if (_.isUndefined(place.name)) {
+    return callback('Name property missing on place.');
+  }
+  if (!_.isString(place.name)) {
+    return callback('Property "name" must be a string.');
+  }
+  if (place.parent) {
+    // validate parents also
+    return validatePlace(place.parent, callback);
+  }
```
garethbowen: Maybe validate that place.parent exists unless this is a district_hospital? I don't think it'd be "correct" to have a clinic without a health_centre for example.

controllers/places.js
```diff
@@ -0,0 +1,111 @@
...
+   */
+  createPlaces: function(place, callback) {
+    var self = this;
+    if (_.isString(place.parent)) {
+      self._getPlace(place.parent, function(err, doc) {
+        if (err) {
+          return callback(err);
+        }
+        place.parent = doc;
+        self.createPlaces(place, callback);
```
garethbowen: I find this recursion quite confusing, although I think it's correct. I think it would be clearer to just call `self._createPlace(place, callback);` directly because that's where it's going to end up anyway.

controllers/places.js
mandric: Yep makes sense. 

We also had a `national_office` facility at one point, not sure what happened to that or if it's still in use.  It was useful theoretically to assign to the national admin role.  Probably for another day, or just another bandaid, but something I've been meaning to mention.

controllers/places.js
mandric: It's goes to `createPlace` only when all the parents are resolved... maybe a couple comments would help.

controllers/users.js
```diff
@@ -232,6 +194,52 @@ var createUserSettings = function(data, response, callback) {
   });
 };
 
+var createPlace = function(data, response, callback) {
```
garethbowen: This feels very similar to places.createPlaces. Maybe it should be in the Places controller and then see if there's some code we can reuse?

controllers/users.js
```diff
@@ -416,13 +425,14 @@ module.exports = {
   _getContactParent: db.medic.get,
```
garethbowen: This is confusing - why not call db.medic.get directly?

mandric: Just future proofs the tests a tiny bit if we need to do something fancier rather than a simple get in the future.

controllers/users.js
mandric: Refactored.

PR #59, opened 2016-04-21
places and contacts support WIP

mandric:
Issue medic/medic-webapp#2046


API_v1.md
```diff
@@ -689,15 +927,15 @@ Create a new user with a place and a contact.
 
 ### Permissions
 
-`can_create_users`  
+`can_create_users`, `cat_create_places`, `cat_create_contacts`
```
garethbowen: typos: s/cat/can

garethbowen: Make sure these permissions are defined in medic-webapp/packages/kujua-sms/views/lib/app_settings.js otherwise there's no easy way to set them.

API_v1.md
```diff
@@ -689,15 +927,15 @@ Create a new user with a place and a contact.
 
 ### Permissions
 
-`can_create_users`  
+`can_create_users`, `can_create_places`, `can_create_contacts`
```
garethbowen: Make sure these permissions are defined in medic-webapp/packages/kujua-sms/views/lib/app_settings.js otherwise there's no easy way to set them.

controllers/contacts.js
```diff
@@ -0,0 +1,116 @@
...
+      return callback(err);
+    }
+    if (doc.type !== 'person') {
+      return error();
+    }
+    callback(null, doc);
+  });
+};
+
+var isAContact = function(obj) {
```
garethbowen: Confusing name - places are contacts too. Maybe call this `isAPerson`?

controllers/contacts.js
```diff
@@ -0,0 +1,116 @@
+var _ = require('underscore'),
+    db = require('../db'),
+    places = require('./places');
+
+var getContact = function(id, callback) {
```
garethbowen: As below - should be `getPerson`

controllers/contacts.js
```diff
@@ -0,0 +1,116 @@
...
+
+/*
+ * Set contact type, validate and write to database. Optionally create place.
+ *
+ * Warning: not doing validation of the contact data against a form yet.  The
+ * form is user defined in settings so being liberal with what gets saved to
+ * the database. Ideally CouchDB would validate a given object against a form
+ * in validate_doc_update. https://github.com/medic/medic-webapp/issues/2203
+ */
+var createContact = function(data, callback) {
```
garethbowen: Rename to `createPerson`

API_v1.md
mandric: Which role should allow these permissions? `national_admin` and `district_admin`?

controllers/places.js
```diff
@@ -111,35 +134,114 @@ var createPlaces = function(place, callback) {
...
+  return place;
+};
+
+var updatePlace = function(id, data, callback) {
+  var self = module.exports,
+      props = PLACE_EDITABLE_FIELDS,
+      response = {},
+      series = [],
+      place;
+  if (!_.some(props, function(k) { return !_.isUndefined(data[k]); })) {
```
garethbowen: Having all this on one line is a bit hard to read.

controllers/places.js
```diff
@@ -111,35 +134,114 @@ var createPlaces = function(place, callback) {
...
+          response.id = resp.id;
+          response.rev = resp.rev;
+          cb();
+        });
       });
-    } else {
-      callback('Place must be a new object or string identifier (UUID).');
-    }
+    });
+    async.series(series, function(cb) {
```
garethbowen: Shouldn't this be: `async.series(series, function(err) {`, ie: "err" not "cb"?

controllers/places.js
```diff
@@ -111,35 +134,114 @@ var createPlaces = function(place, callback) {
...
+      if (err) {
+        return cb(err);
```
garethbowen: And this should be `callback(err)`, right?

controllers/places.js
```diff
@@ -111,35 +134,114 @@ var createPlaces = function(place, callback) {
...
+  } else if (_.isObject(place) && _.isUndefined(place._id)) {
+    // create and return place
+    self._createPlaces(place, function(err, resp) {
+      if (err) {
+        return callback(err);
+      }
+      self.getPlace(resp.id, callback);
+    });
+  } else {
+    callback('Place must be a new object or string identifier (UUID).');
```
garethbowen: Always wrap errors in `new Error('blah')` so we get the stacktrace.

server.js
```diff
@@ -453,6 +465,56 @@ app.delete('/api/v1/users/:username', jsonParser, function(req, res) {
...
+      if (err) {
+        return serverUtils.error(err, req, res);
+      }
+      res.json(body);
+    });
+  });
+});
+
+app.post('/api/v1/contacts', jsonParser, function(req, res) {
+  auth.check(req, 'can_create_contacts', null, function(err) {
```
garethbowen: I think this endpoint should be `/api/v1/people/` and the permission should be `can_create_people`.

server.js
```diff
@@ -453,6 +465,56 @@ app.delete('/api/v1/users/:username', jsonParser, function(req, res) {
...
+    if (err) {
+      return serverUtils.error(err, req, res);
+    }
+    if (_.isEmpty(req.body)) {
+      return emptyJSONBodyError(req, res);
+    }
+    contacts.createContact(req.body, function(err, body) {
```
garethbowen: `people.createPerson`

API_v1.md
garethbowen: Probably by default anyone can create places and people, right? That's the way the webapp works currently, and there's no reason why they should be able to do it there and not through the API.

mandric:
@gareth I updated contact -> person in the API doc, except in the context of places, where it sounds better and was part of the underlying document structure and part of the API request too, because we set `contact` on places not `person`.  So I continued using "contact" when describing people in relation to places, have a sanity check when you have a moment.  Thanks!

Examples:

"Update a place's contact." is better than "Update a place's person."

"Create a new user with a place and a contact." is better than "Create a new user with a place and a person."

"Update a place and optionally its contact." is better than "Update a place and optionally its person."

If the doc makes sense to you then I will continue updating the code to follow along.

garethbowen:
@mandric Very nice. Go for it!

mandric:
@garethbowen should be much better now.

mandric:
Just came across one thing I should do though before we merge this.  A simple check (head request behind the scenes) in users api for a conflict with user or user-settings doc before creating the required places/contacts.  This will avoid having duplicate places and contacts when you re-run bulk loads using the /api/v1/users API.

PR #69, opened 2016-05-25
Add integration with `medic-gateway` for sending and receiving SMS

alxndrsn:


controllers/sms-gateway.js
```diff
@@ -0,0 +1,186 @@
...
+      }
+      return body;
+    });
+}
+
+function saveToDb(gatewayRequest, wtMessage) {
+  var messageBody = querystring.stringify({
+    from: wtMessage.from,
+    message: wtMessage.content,
+    'medic-gateway_id': wtMessage.id,
```
mandric: Is this the id message or gateway message? I wonder if its data about the gateway software or the device.

controllers/sms-gateway.js
```diff
@@ -0,0 +1,186 @@
...
+          .catch(reject);
+      });
+
+    req.on('error', reject);
+
+    req.write(messageBody);
+    req.end();
+  })
+  .catch(function(err) {
+    console.log('sms-gateway', 'saveToDb', 'error saving', wtMessage, err);
```
mandric: Maybe console.error here is better? Or server-utils.error()?

controllers/sms-gateway.js
```diff
@@ -0,0 +1,186 @@
...
+}
+
+function updateStateFor(gatewayRequest, update) {
+  var newState = getWebappState(update);
+  if (!newState) {
+    return Promise.reject(new Error('Could not work out new state for update: ' + JSON.stringify(update)));
+  }
+
+  return updateState(
+      gatewayRequest,
```
mandric: Spacing

controllers/sms-gateway.js
```diff
@@ -0,0 +1,186 @@
...
+      gatewayRequest.headers['user-agent'],
+      update.id,
+      newState);
+}
+
+function updateState(gatewayRequest, userAgent, messageId, newState) {
+  var updateBody = {
+    state: newState,
+    details: {
+      useragent: userAgent,
```
mandric: I'm not sure we need userAgent in every state update? Bloats the doc? I think userAgent is only necessary in the http header. And reserve the details property for information about the state chage, like "out of credit" or something.

mandric: Actually maybe the device ID might be useful. Not sure now.

controllers/sms-gateway.js
```diff
@@ -0,0 +1,186 @@
...
+  new Promise(function(resolve, reject) {
+    messageUtils.updateMessage(messageId, updateBody, UNUSED, function(err, result) {
+      if (err) {
+        return reject(err);
+      }
+      return resolve();
+    });
+  })
+  .catch(function(err) {
+    console.log('sms-gateway', 'updateState', 'error updating message state', userAgent, messageId, newState, err);
```
mandric: I think we want a stacktrace  in the logs here. Just passing this to.server-utils.error might be best or just throw?

mandric: Sorry server-utils.error is only in the develop branch but I think you get what I mean. For these types of unhandled errors we want stack traces in the logs. You might log the arguments and then throw the error? Or is that effectively the same thing? I think we want to at least avoid `console.log` for logging errors have it handled centrally like in the develop branch.

controllers/sms-gateway.js
alxndrsn: It's the id assigned to the message by `medic-gateway` - obviously needs a better name!

controllers/sms-gateway.js
alxndrsn: spacing looks ok to me...

controllers/sms-gateway.js
alxndrsn: Me neither.  Can't see it being very useful in our normal use case though, so maybe better of leaving it out.  I'm not sure if leaving `details` undefined is valid though.

controllers/sms-gateway.js
alxndrsn: :+1: i will backport the useful bit of server-utils.js to `master` and use the logging from there.

controllers/sms-gateway.js
alxndrsn: How about one of:
* `tracking_id`
* `tracking_ref`
* `gateway_ref`

mandric: Why not message_id?

controllers/sms-gateway.js
mandric: It's 2 everywhere but 4 here. But if you like it then keep it.

controllers/sms-gateway.js
mandric: I would put in a device_id or leave it undefined. User agent string won't be useful... but device id could be useful to track down what gateway device made what changes.

mandric: Or what device is out of credit.

mandric: Actually this was originally created for the reason and other details. So you could use it for that if you want or ignore it.

controllers/sms-gateway.js
```diff
@@ -0,0 +1,187 @@
...
+    message: wtMessage.content,
+    'medic-gateway_id': wtMessage.id,
+  });
+
+  return new Promise(function(resolve, reject) {
+    var req = http.request(
+      {
+        hostname: 'localhost',
+        port: 5984,
+        path: '/medic/_design/medic/_rewrite/add',
```
mandric: I'd prefer if you used /api/v1/records if possible. Either via http or require the controller... we can open an issue for refactoring common controller bits.

mandric: The less references we have to direct couch api the better.

controllers/sms-gateway.js
alxndrsn: Because the object also includes:
```{
  "type": "message",
  "_id": ...
}```
so adding a property `message_id` doesn't seem like it would be very clear...

controllers/sms-gateway.js
alxndrsn: This stuff is double-indented as it's a continuation of the line above, rather than a new block.  Seems like the standard convention to me.

controllers/sms-gateway.js
alxndrsn: Yeah i will ditch it for now - don't know of any current projects which have multiple gateway phones.

controllers/sms-gateway.js
mandric: Ahh ok

controllers/sms-gateway.js
mandric: Thoughts?

alxndrsn: Would this involve copy pasting the code for `/add` into this controller?

mandric: Of course not, see previous comment. Either POST /api/v1/records or require('./records').createRecord() etc...
"I'd prefer if you used /api/v1/records if possible. Either via http or require the controller... we can open an issue for refactoring common controller bits."

mandric: It's not a big deal but since we have /api/v1/records it makes sense to use it!

mandric: If it's a pain for you just create an issue for it and we can deal with it later.  Feel free merge.

controllers/sms-gateway.js
alxndrsn: Any of those name options look better to you?

controllers/sms-gateway.js
alxndrsn: Should be fixed

alxndrsn:
@mandric I think I've addressed your concerns.  I will continue writing more tests, and then merge.

PR #70, opened 2016-05-25
remove district arg from message controller functions

alxndrsn:
This arg doesn't seem to do anything.

### DIFF
--- a/controllers/messages.js
+++ b/controllers/messages.js
@@ -2,7 +2,7 @@ var db = require('../db-nano');
 
 module.exports = {
 
-  getMessages: function(options, district, callback) {
+  getMessages: function(options, callback) {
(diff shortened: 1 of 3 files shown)


mandric:
Anyone have a clue why that was there in the first place? That seems too weird.

mandric:
Only thing I can think of is since we do authorization based on district. Looks like a placeholder.

alxndrsn:
Seems to be equally pointless in `develop`.  @garethbowen any objections to merging this and applhing the same fix to `develop`?

garethbowen:
Yeah that looks like a hangover from the old days when restricted users could access couch directly. Fire when ready.

alxndrsn:
TODO: merge to `develop`

PR #78, opened 2016-06-14
added support for `reported_date` in people and places APIs.

mandric:
Issue: https://github.com/medic/medic-webapp/issues/2046


mandric:
Some duplication in here that I should probably move to test in utils or something.

PR #81, opened 2016-06-16
New docs should replicate. Issue: medic/medic-webapp#2396

garethbowen:
This is becoming quite complicated but it's still the best solution I can think of, at least in the short term.

### DIFF
--- a/handlers/changes.js
+++ b/handlers/changes.js
@@ -6,6 +6,10 @@ var _ = require('underscore'),
     ALL_KEY = [ '_all' ], // key in the doc_by_place view for records everyone can access
     UNASSIGNED_KEY = [ '_unassigned' ]; // key in the doc_by_place view for unassigned records
 
+var error = function(code, message) {
+  return JSON.stringify({ code: code, message: message });
+};
+
 var getViewKeys = function(req, userCtx, callback) {
   auth.getFacilityId(req, userCtx, function(err, facilityId) {
     if (err) {
@@ -58,7 +62,7 @@ var prepareResponse = function(req, res, changes, verifiedIds) {
   });
   if (rejected.length) {
     console.error('User attempting to replicate these docs without permission: ', _.pluck(rejected, 'id'));
-    return res.write(JSON.stringify({ error: 'Forbidden' }));
+    return res.write(error(403, 'Forbidden'));
   }
   res.write(JSON.stringify(changes));
 };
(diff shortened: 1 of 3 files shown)


handlers/changes.js
```diff
@@ -16,112 +22,238 @@ var getViewKeys = function(req, userCtx, callback) {
...
+  bindRequestedIds(feed, function(err) {
+    if (err) {
+      return callback(err);
+    }
+    bindViewKeys(feed, function(err) {
+      if (err) {
+        return callback(err);
+      }
+      bindValidatedDocIds(feed, callback);
+    });
```
SCdF: I have no problem with this and its called functions, but aside question about callbacks vs. promises vs javascript: Are promises always better than callbacks, or are callbacks sometimes more appropriate? Is there some idiomatic rule in Javascript. or have you just decided to use callbacks here because it makes more sense in relation to the code around it?

handlers/changes.js
```diff
@@ -16,112 +22,238 @@ var getViewKeys = function(req, userCtx, callback) {
...
-    return res.write(JSON.stringify({ error: 'Forbidden' }));
-  }
-  res.write(JSON.stringify(changes));
 };
 
-var getRequestIds = function(req, callback) {
-  if (req.body && req.body.doc_ids) {
-    // POST request
-    return callback(null, req.body.doc_ids);
+// return true if the user should be able to see this doc AND they can't already
```
SCdF: This comment is a bit misleading? AFAICT it's more like: `// returns if it is true that for any document in the feed, the user should be able to see it AND they don't already`

handlers/changes.js
```diff
@@ -16,112 +22,238 @@ var getViewKeys = function(req, userCtx, callback) {
...
+var isNewApplicableDoc = function(feed, docs) {
+  return _.some(docs, function(doc) {
+    return !_.contains(feed.validatedIds, doc.id) &&
+      _.some(doc.keys, function(key) {
+        return !!_.find(feed.keys, function(feedKey) {
```
SCdF: Well I learnt another piece of Javascript magic today, so that's something.

handlers/changes.js
```diff
@@ -16,112 +22,238 @@ var getViewKeys = function(req, userCtx, callback) {
...
-  }
-  res.write(JSON.stringify(changes));
 };
 
-var getRequestIds = function(req, callback) {
-  if (req.body && req.body.doc_ids) {
-    // POST request
-    return callback(null, req.body.doc_ids);
+// return true if the user should be able to see this doc AND they can't already
+var isNewApplicableDoc = function(feed, docs) {
```
SCdF: Maybe rename to `hasNewApplicableDoc` since it's checking many docs?

handlers/changes.js
garethbowen: The callbacks vs promises argument rages on. In most cases they are completely equivalent. There is one special case where callbacks are required (multiple invocations) and one case when promises are much nicer (asynchronous initialisation), but other than that it's personal preference. My personal preference is to use promises.

However, it's much nicer to use one or the other as much as possible within one codebase because switching between is buggy and ugly (buggly?). In API we use callbacks everywhere so I stuck with it here for consistency. I'm open to the idea of refactoring API to use promises.

handlers/changes.js
garethbowen: Good call. I think at some point I must've refactored it to check multiple docs.

PR #82, opened 2016-06-28
Remove translations from ddoc. Issue: medic/medic-webapp#1706

garethbowen:


config.js
```diff
@@ -64,62 +63,74 @@ var getMessage = function(value, locale) {
...
-    db.medic.get('_design/medic', function(err, ddoc) {
-      if (err) {
-        return callback(err);
-      }
-      settings = ddoc.app_settings;
-      _.defaults(settings, defaults);
-      callback();
-    });
+    loadSettings(callback);
+    loadTranslations();
```
mandric: Any particular reason you didn't do loadTranslations after loadSettings?  `load` is calling the callback without really being done?

mandric:
```
loadSettings(function(err) {
  if (err) callback(err);
  loadTranslations(callback);
});
```

garethbowen: I'm happy with translations loading in the background. We can't callback until the settings are loaded because of permissions and so on, but I'm happy for translations to be loaded just as soon as possible without blocking.

PR #84, opened 2016-07-01
allow `reported_date` on places/people to use MS since Unix Epoch.

mandric:
This is also the stored format (ms since epoch) so we need to support it
in the validation anyway otherwise validating existing records will
fail.

Also added a bit of centralized documentation about timestamps.  The
other record types (Records) need to be updated still to support the new
format and that will get done soon.

https://github.com/medic/medic-webapp/issues/2449


mandric:
Just realized I need to add some parse tests here.

mandric:
Ready

PR #86, opened 2016-07-07
Client side ddoc. Issue: medic/medic-webapp#2206

garethbowen:


config.js
```diff
@@ -76,7 +77,7 @@ var loadSettings = function(callback) {
 
 var loadTranslations = function() {
   var options = { key: [ 'translations', true ], include_docs: true };
-  db.medic.view('medic', 'doc_by_type', options, function(err, result) {
+  db.medic.view('medic-client', 'doc_by_type', options, function(err, result) {
```
alxndrsn: Again this looks like a non-client using the client-specific ddoc.  But again, I may be misunderstanding the use of the word "client" in this context.

config.js
```diff
@@ -126,6 +127,11 @@ module.exports = {
             process.exit(1);
           }
         });
+        ddocExtraction.run(function(err) {
```
alxndrsn: This feels like maybe it should be more of a fatal error...

server.js
```diff
@@ -593,34 +593,34 @@ proxyForAuditing.on('error', function(err, req, res) {
...
+      console.log('Medic API listening on port ' + apiPort);
+    });
+    translations.run(function(err) {
       if (err) {
-        return console.error(err);
+        return console.error('Error merging translations', err);
       }
-      console.log('Database migrations completed successfully');
+      console.log('Translations merged successfully');
+      migrations.run(function(err) {
```
alxndrsn: ...I guess this answers my migrations vs ddoc extraction race question

alxndrsn:
LGTM

PR #89, opened 2016-07-20
Errors no longer error medic/medic-webapp#2520

SCdF:
**HELLO REVIEWER,** I put this in a PR because I'm not sure on our stance on changing migrations once they're live. This seems like a strict upgrade, but I thought I'd check anyway.

This happened because the errors created attempt to reference the
object that failed to be loaded and caused the error to be created
in the first place.

(it's really hard to explain that concisely)

### DIFF
--- a/migrations/extract-person-contacts.js
+++ b/migrations/extract-person-contacts.js
@@ -113,7 +113,7 @@ var createPerson = function(id, callback) {
   db.medic.get(id, function(err, facility) {
     if (err) {
       if (err.statusCode === 404) {
-        return callback(new Error('facility ' + facility._id + ' not found.'));
+        return callback(new Error('facility ' + id + ' not found.'));
       }
       return callback(err);
     }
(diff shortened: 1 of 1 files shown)


alxndrsn:
Definitely seems like the fix is good for future projects which will run this migration, but what is the effect for projects which already have this migration deployed?  Will their migrations have broken part-way through?  Will re-running this migration with the fix cause any problems with data integrity?

SCdF:
> Will their migrations have broken part-way through? 

I don't think so, as I believe this error causes the migrations to error out and fail, so that means they won't be marked completed?

SCdF:
@garethbowen what are your thoughts on this one? Cool to merge?

garethbowen:
Yeah my view is that it's ok to modify a migration if you're fixing a fatal bug like this because anyone who would've hit it hasn't completed the migration so when they update to the next version it'll run as expected. It's certainly not ok to add features to a migration.

PR #92, opened 2016-08-08
Do not break extract-person-migration if parent does not exist

alxndrsn:
Issue: #2586


SCdF:
@alxndrsn that test fail looks pretty important: the test is not finishing on Travis

PR #93, opened 2016-08-08
2x sms gateway. Issue: medic/medic-gateway#2530

garethbowen:


controllers/records.js
```diff
@@ -22,13 +22,11 @@ var request = function(opts, callback) {
 var createByForm = function(data, callback) {
   var required = ['message', 'from'],
       optional = ['reported_date', 'locale', 'gateway_ref'];
-  for (var k in required) {
-    if (required.hasOwnProperty(k)) {
-      if (!exists(data[required[k]])) {
```
alxndrsn: This doesn't look like it should have worked before.

controllers/records.js
```diff
@@ -22,13 +22,11 @@ var request = function(opts, callback) {
...
-        return callback(new Error('Missing required field: ' + required[k]));
-      }
+  required.forEach(function(key) {
+    if (!exists(data[key])) {
+      return callback(new Error('Missing required field: ' + required[k]));
```
alxndrsn: This error message might should be `'Missing required field: ' + k`.

controllers/records.js
```diff
@@ -44,20 +42,20 @@ var createRecordByJSON = function(data, callback) {
...
     return callback(new Error('Missing _meta property.'));
   }
-  for (var k in required) {
-    if (required.hasOwnProperty(k)) {
-      if (!exists(data._meta[required[k]])) {
-        return callback(new Error('Missing required field: ' + required[k]));
-      }
+  required.forEach(function(key) {
+    if (!exists(data._meta[key])) {
+      return callback(new Error('Missing required field: ' + required[k]));
```
alxndrsn: Again this error message now looks wrong - `k` is not defined.

controllers/sms-gateway.js
```diff
@@ -119,6 +120,7 @@ module.exports = {
       .then(JSON.parse)
       .then(function(request) {
         // Process webapp-terminating messages asynchronously
+        // TODO these promises do nothing!
```
alxndrsn: This comment is unclear - it looks like these promises make the contained code asynchronous, but is the implication that this can be simplified?

controllers/sms-gateway.js
```diff
@@ -55,103 +34,111 @@ function getWebappState(update) {
...
-  if (update.status === 'FAILED' && update.reason) {
-    return updateState(update.id, newState, update.reason);
-  }
-
-  return updateState(update.id, newState);
+  updateState(update.id, newState, update.reason, callback);
 }
 
-function updateState(messageId, newState, failureReason) {
+function updateState(messageId, newState, reason, callback) {
```
alxndrsn: I think `failureReason` makes the use of the `reason` field more explicit.

controllers/records.js
```diff
@@ -22,11 +22,13 @@ var request = function(opts, callback) {
 var createByForm = function(data, callback) {
   var required = ['message', 'from'],
       optional = ['reported_date', 'locale', 'gateway_ref'];
-  required.forEach(function(key) {
-    if (!exists(data[key])) {
-      return callback(new Error('Missing required field: ' + required[k]));
+  for (var k in required) {
```
alxndrsn: Doesn't look like the correct way to iterate over an array.

controllers/records.js
```diff
@@ -42,11 +44,13 @@ var createRecordByJSON = function(data, callback) {
   if (!exists(data._meta)) {
     return callback(new Error('Missing _meta property.'));
   }
-  required.forEach(function(key) {
-    if (!exists(data._meta[key])) {
-      return callback(new Error('Missing required field: ' + required[k]));
+  for (var k in required) {
```
alxndrsn: Doesn't look like the correct way to iterate over an array.

alxndrsn:
Comments inline.

controllers/sms-gateway.js
garethbowen: They weren't making the code any more asynchronous and a loop would achieve the same result. This was a WIP commit and the TODO is now TODONE.

controllers/records.js
garethbowen: Yeah this commit broke. It was picked up in tests and I fixed it in a later commit in this PR.

controllers/sms-gateway.js
alxndrsn: The point of the promises was to make the processing of messages happen after the response has been sent to the client, and the connection closed.  I appreciate they may not have been doing this.

garethbowen: I get that, but because `saveToDb` is async already all this `Promise.resolve()` is doing is providing a second async layer. If you just have the loop which fires off all the `saveToDb` calls none of them will complete (or even start) before this function has returned.

alxndrsn: 👍 Glad that's covered.  I also could have done a better job documenting the intention of this in the first place.

PR #95, opened 2016-08-10
Add migration to remove empty parents from contacts

alxndrsn:
Issue: medic/medic-webapp#2436

### DIFF
--- a/migrations/remove-empty-parents.js
+++ b/migrations/remove-empty-parents.js
@@ -1,0 +1,55 @@
+var async = require('async'),
+    db = require('../db');
+
+function hasProperties(obj) {
+  var k;
+
+  if (!obj) {
+    return false;
+  }
+
+  for(k in obj) {
+    if(obj.hasOwnProperty(k)) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+function removeDeadParents(row, callback) {
+  console.log('Should process parent: ', row);
+  if(hasProperties(row.doc.parent)) {
+    return callback();
+  }
+  db.medic.get(row.id, function(err, doc) {
+    if(err) {
+      return callback(err);
+    }
+    delete doc.parent;
+    db.medic.insert(doc, function(err) {
+      if(err) {
+        return callback(err);
+      }
+      return callback();
+    });
+  });
+}
+
+module.exports = {
+  name: 'remove-empty-parents',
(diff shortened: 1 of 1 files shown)


migrations/remove-empty-parents.js
```diff
@@ -0,0 +1,55 @@
...
+    return false;
+  }
+
+  for(k in obj) {
+    if(obj.hasOwnProperty(k)) {
+      return true;
+    }
+  }
+
+  return false;
```
garethbowen: What about `return obj && !_.isEmpty(obj)`?

migrations/remove-empty-parents.js
```diff
@@ -0,0 +1,55 @@
...
+  console.log('Should process parent: ', row);
+  if(hasProperties(row.doc.parent)) {
+    return callback();
+  }
+  db.medic.get(row.id, function(err, doc) {
+    if(err) {
+      return callback(err);
+    }
+    delete doc.parent;
+    db.medic.insert(doc, function(err) {
```
garethbowen: It looks like you can just pass the `callback` instead of a new anonymous function.

migrations/remove-empty-parents.js
```diff
@@ -0,0 +1,55 @@
...
+    if(obj.hasOwnProperty(k)) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+function removeDeadParents(row, callback) {
+  console.log('Should process parent: ', row);
```
garethbowen: This seems a bit spammy.

migrations/remove-empty-parents.js
```diff
@@ -0,0 +1,55 @@
...
+      return callback();
+    });
+  });
+}
+
+module.exports = {
+  name: 'remove-empty-parents',
+  created: new Date(2016, 7, 10, 13, 37, 0, 0),
+  run: function(callback) {
+    // TODO pull full list of contacts, and remove dead parents
```
garethbowen: I think this TODO is done?

migrations/remove-empty-parents.js
alxndrsn: A mistake!

migrations/remove-empty-parents.js
alxndrsn: Oh yes it is!

PR #99, opened 2016-09-02
Streaming Exports

SCdF:
Issue #2694

Exports now stream from API instead of being generated as one giant string, reducing memory requirements.

This is a POC that only converts `outputToXml`. It only streams XML generation, not pulling data from CouchDB. This would be the next natural step if we still get memory issues.

I'll work on converting the other output formats to stream as well.

### DIFF
--- a/controllers/export-data.js
+++ b/controllers/export-data.js
@@ -449,14 +449,18 @@ var outputToCsv = function(options, tabs, callback) {
 };
 
 var outputToXml = function(options, tabs, callback) {
+  var chunkier = '';
+  var concat = function(chunk) {chunkier = chunkier + chunk;};
 
-  var workbook = xmlbuilder.create('Workbook', { encoding: 'UTF-8' }, {}, { allowSurrogateChars: true })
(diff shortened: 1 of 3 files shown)


controllers/export-data.js
```diff
@@ -449,38 +449,51 @@ var outputToCsv = function(options, tabs, callback) {
...
+    tabs.forEach(function(tab) {
+      var table = workbook
+        .ele('Worksheet', { 'ss:Name': tab.name })
+        .ele('Table');
+
+      if (!options.skipHeader) {
+        row = table.ele('Row');
+        tab.columns.forEach(function(column) {
+          row.ele('Cell').ele('Data', {'ss:Type': 'String'}, column.label);
+          row.up().up();
```
SCdF: Main additions here is the use of `up()` to move the 'pointer' back to the correct place. This is required for streaming as opposed to keeping the whole thing in memory, otherwise invalid XML is generated.

mandric:
Awesome!

mandric:
The CSV output should be streaming already I think...

PR #100, opened 2016-11-01
Make audit perform by using alldocs instead of a view. Issue: medic/medic-webapp#2879

garethbowen:


migrations/drop-audit-doc-index.js
```diff
@@ -22,7 +22,7 @@ var needsUpdate = function(row) {
 var createNewDocs = function(oldDocs, callback) {
   var newDocs = oldDocs.map(function(doc) {
     return {
-      _id: doc.record_id,
+      _id: doc.record_id + '-audit',
```
browndav: Is this better structured as a prefix? Not a strong preference, but IIRC CouchDB uses prefixes as a convention and they probably have some highly-theoretical future benefits wrt indexing.

migrations/drop-audit-doc-index.js
```diff
@@ -0,0 +1,115 @@
...
+    createNewDocs(oldDocs, function(err) {
+      if (err) {
+        return callback(err);
+      }
+      deleteOldDocs(oldDocs, function(err) {
+        if (err) {
+          return callback(err);
+        }
+        // The newSkip is based on the old skip, plus the number of
+        // unchanged files. Changed files can't be included because
```
browndav: 'Files' reads a little awkwardly here; just to be sure, we're talking about docs/records, right?

migrations/drop-audit-doc-index.js
```diff
@@ -0,0 +1,115 @@
...
+        // they have just been deleted so will not be included in the
+        // next query result. This means some records will be processed
+        // more than once but gaurantees every record will be processed
+        // at least once.
+        var newSkip = skip + BATCH_SIZE - oldDocs.length;
```
browndav: If this encounters a run of 100 records, all of which are missing a `record_id` or have a type other than `audit_record`, it looks like this would loop infinitely. If I'm reading this right, I'd probably suggest that we enforce some kind of forward progress (e.g. `skip + max(BATCH_SIZE - oldDocs.length, 1)`).

garethbowen: I don't think that's right. `oldDocs` are all updated, so while you're right that if `oldDocs.length === BATCH_SIZE` then `newSkip` will be the same as `skip`, however on the next run through `oldDocs` will be different and `skip` will change. This is the part of the algorithm I'm most concerned about so it's definitely worth another look.

migrations/drop-audit-doc-index.js
browndav: Yeah, highly theoretical. Beside is better; +1.

migrations/drop-audit-doc-index.js
browndav: Yeah, looks like I missed that the side-effect of actually updating records matters to `oldDocs`. Probably worth a comment, but I agree with your analysis.

PR #101, opened 2016-11-08
Refining auditing hooks medic/medic-webapp#2895

SCdF:
- No trailing slash detection has been improved to include standard
   db path
 - All of the CouchDB API that uses POST but doesn't write (up to 2.0)
   is no longer audited

### DIFF
--- a/server.js
+++ b/server.js
@@ -85,21 +85,28 @@ app.use(express.static(path.join(__dirname, 'public')));
 app.get(pathPrefix + 'login', login.get);
 app.post(pathPrefix + 'login', jsonParser, login.post);
 
-app.all(appPrefix + 'update_settings/*', function(req, res) {
-  // don't audit the app settings
-  proxy.web(req, res);
-});
-app.all(pathPrefix + '_revs_diff', function(req, res) {
-  // don't audit the _revs_diff
-  proxy.web(req, res);
-});
-app.all(pathPrefix + '_local/*', function(req, res) {
-  // don't audit the _local docs
-  proxy.web(req, res);
-});
-app.all(pathPrefix + '_bulk_get/*', function(req, res) {
-  // don't audit the _bulk_get request
-  proxy.web(req, res);
+// Requests we don't wish to audit
+// (usually because POST is just used here to pass more parameters)
+[
+  appPrefix + 'update_settings/*',
+  // Replication machinery we don't care to audit
+  pathPrefix + '_local/*',
+  pathPrefix + '_revs_diff',
+  pathPrefix + '_missing_revs',
+  // Can use POST for specifiying ids
+  pathPrefix + '_all_docs',
+  pathPrefix + '_design/*/_list/**',
+  pathPrefix + '_design/*/_show/**',
+  pathPrefix + '_design/*/_view/*',
+  pathPrefix + '_changes',
+  // Interacting with mongo filters uses POST
+  pathPrefix + '_find',
+  pathPrefix + '_explain',
+
+].forEach(function(url) {
+  app.all(url, function(req, res) {
+    proxy.web(req, res);
+  });
 });
 
 app.get('/setup/poll', function(req, res) {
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -85,21 +85,28 @@ app.use(express.static(path.join(__dirname, 'public')));
...
-});
-app.all(pathPrefix + '_revs_diff', function(req, res) {
-  // don't audit the _revs_diff
-  proxy.web(req, res);
-});
-app.all(pathPrefix + '_local/*', function(req, res) {
-  // don't audit the _local docs
-  proxy.web(req, res);
-});
-app.all(pathPrefix + '_bulk_get/*', function(req, res) {
```
SCdF: NB: Deleted this as it isn't in the CouchDB api (it looks to have been in Cloudant at at least one point) and we don't use it anywhere

server.js
```diff
@@ -85,21 +85,28 @@ app.use(express.static(path.join(__dirname, 'public')));
...
-  // don't audit the _bulk_get request
-  proxy.web(req, res);
+// Requests we don't wish to audit
+// (usually because POST is just used here to pass more parameters)
+[
+  appPrefix + 'update_settings/*',
```
SCdF: @garethbowen do you know why we don't audit changes to settings? Is it because we know only admins can do it, or is this the standard "GETting using POST" issue?

server.js
```diff
@@ -598,8 +605,13 @@ proxy.on('proxyReq', function(proxyReq, req, res) {
 /**
  * Redirect to add the trailing slash without which mysterious bugs occur...
  */
-app.get(pathPrefix + '_design/' + db.settings.ddoc + '/_rewrite', function(req, res) {
-  res.redirect(appPrefix);
+[
+  appPrefix,
+  pathPrefix
+].forEach(function(url) {
+  app.get(url.substring(url.length - 1), function(req, res) {
```
SCdF: I'm not completely comfortable with this. It:
 - Presumes that anything added to that list has a `/` at the end (which for now it does)
 - And then uses that knowledge to remove it from the end

I like it more than what was there before though, where we effectively define the variable twice, and opposite ends of the file, and have one have a trailing space and the other not.

server.js
SCdF: Whoops I get it: PouchDB tries to use it. I'll add it back in.

server.js
garethbowen: Auditing works by taking the given doc so we can get the ID and therefore find the audit doc later. The update_settings endpoint doesn't take a doc it just takes a JSON string with no ID. We could audit it but we'd need to think about how to do that (eg: inject the right ID?). But the bottom line is the existing audit framework can't handle objects without IDs.

server.js
garethbowen: You could use charAt to check the last character is a slash? Also, using `slice` is a little prettier, eg: `url.slice(0, -1)`

server.js
```diff
@@ -85,21 +85,28 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+[
+  appPrefix + 'update_settings/*',
+  // Replication machinery we don't care to audit
+  pathPrefix + '_local/*',
+  pathPrefix + '_revs_diff',
+  pathPrefix + '_missing_revs',
+  // Can use POST for specifiying ids
+  pathPrefix + '_bulk_get',
+  pathPrefix + '_all_docs',
+  pathPrefix + '_design/*/_list/**',
```
garethbowen: I think `**` is equivalent to `*` isn't it?

server.js
```diff
@@ -85,21 +85,28 @@ app.use(express.static(path.join(__dirname, 'public')));
...
+  pathPrefix + '_design/*/_show/**',
+  pathPrefix + '_design/*/_view/*',
+  pathPrefix + '_changes',
+  // Interacting with mongo filters uses POST
+  pathPrefix + '_find',
+  pathPrefix + '_explain'
+].forEach(function(url) {
```
garethbowen: I think it'd be clearer to pull this array out as a constant, something like "UNAUDITED_ENDPOINTS", idk. Then while iterating over them add the `pathPrefix`.

server.js
SCdF: Weird, I don't think it's in the couchdb2.0 docs, since that's what I used. In any case, I added it back.

server.js
SCdF: In all honesty I entered that instinctively based on other frameworks I've used. I'll check Node's documentation on that one.

server.js
garethbowen: You're right - it's missing from the documentation. Here's the closest thing I could find: http://developer.couchbase.com/documentation/mobile/1.1.0/develop/references/sync-gateway/rest-api/database/post-bulk-get/index.html

PR #105, opened 2016-11-17
Version check medic/medic-webapp#2931 medic/medic-webapp#2854

SCdF:
Wow that diffs badly doesn't it. I only wrapped and added the first check

### DIFF
--- a/server.js
+++ b/server.js
@@ -650,34 +650,56 @@ proxyForAuditing.on('error', function(err, req, res) {
   serverUtils.serverError(JSON.stringify(err), req, res);
 });
 
-ddocExtraction.run(function(err) {
+var couchdbCheck = function(callback) {
+  db.request({}, function(err, body) {
+    if (err) {
+      callback(err);
+    }
+
+    var semvers = body.version && body.version.match(/(\d+)\.\d+\.\d+/);
+    if (!semvers || semvers[1] !== '2') {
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -650,34 +650,56 @@ proxyForAuditing.on('error', function(err, req, res) {
...
-ddocExtraction.run(function(err) {
+var couchdbCheck = function(callback) {
+  db.request({}, function(err, body) {
+    if (err) {
+      callback(err);
+    }
+
+    var semvers = body.version && body.version.match(/(\d+)\.\d+\.\d+/);
+    if (!semvers || semvers[1] !== '2') {
+      callback('Only CouchDB 2.x is supported, v' + body.version + ' detected.');
```
garethbowen: Make this `callback(new Error('Only...'));` so we get a stacktrace.

server.js
```diff
@@ -650,34 +650,56 @@ proxyForAuditing.on('error', function(err, req, res) {
...
+    }
+
+    callback();
+  });
+};
+
+couchdbCheck(function(err) {
```
garethbowen: This is too deeply nested now. It was too deeply nested before, but now it's just crazy! Maybe use async.series or something.

SCdF: It's funny, I actually wrote it with promises, then realised "whoops, wrong standard" and changed it to this. I will try to clean it up.

server.js
SCdF: rgr that

PR #106, opened 2016-11-22
Remove usage of the _config endpoint

SCdF:


controllers/users.js
```diff
@@ -34,11 +34,11 @@ var error400 = function(msg, callback) {
   });
 };
 
-var getType = function(user, admins) {
+var getType = function(user) {
   if (user.roles && user.roles.length) {
     return user.roles[0];
   }
-  return admins[user.name] ? 'admin' : 'unknown';
```
SCdF: So again, this is probably controversial, but is sort of the keystone of this PR. Previously we have code-- either because we needed it or because we were being extra cautious-- which checked for CouchDB admins and injected the admin role into results. It's very unclear if we rely on this, and at least for LG admin users actually already have the `_admin` role.

@garethbowen we'll almost certainly have to talk about this. I'm very much getting the feeling that I'm pulling at some pretty long threads here.

tests/unit/controllers/users.js
```diff
@@ -143,18 +138,6 @@ exports['getType returns unknown when roles is empty'] = function(test) {
...
-  var user = {
-    name: 'sam',
-    roles: []
-  };
-  var admins = {
-    'sam': 'x'
-  };
-  test.equal(controller._getType(user, admins), 'admin');
-  test.done();
-};
```
SCdF: This is the unit test we're gonna have to talk about re: the comment above. Since we don't want people using CouchDB admin users we don't really want to support this kind of thing. Question is: what is it enabling (ideally nothing, but you'll know more than me)

tests/unit/controllers/users.js
```diff
@@ -366,57 +344,6 @@ exports['getList handles minimal users'] = function(test) {
...
-    test.equal(err, null);
-    test.equal(data.length, 1);
-    var gareth = data[0];
-    test.equal(gareth.id, 'org.couchdb.user:gareth');
-    test.equal(gareth.username, 'gareth');
-    test.equal(gareth.type, 'national-admin');
-    test.done();
-  });
-};
-
```
SCdF: Seem to be variations on the test above, same delete reason.

tests/unit/controllers/users.js
```diff
@@ -461,40 +388,6 @@ exports['getList returns errors from facilities service'] = function(test) {
...
-    }
-  ]);
-  sinon.stub(controller, '_getAllUserSettings').callsArgWith(0, null, []);
-  sinon.stub(controller, '_getAdmins').callsArgWith(0, 'POW');
-  controller.getList(function(err) {
-    test.equal(err, 'POW');
-    test.done();
-  });
-};
-
```
SCdF: No admin service, no test

tests/unit/controllers/users.js
```diff
@@ -1099,27 +990,6 @@ exports['updateUser updates password on user doc'] = function(test) {
...
-  var update = sinon.stub(controller, '_updateUser').callsArg(2);
-  var updateSettings = sinon.stub(controller, '_updateUserSettings').callsArg(2);
-  controller.updateUser('paul', data, function(err) {
-    test.ok(!err);
-    test.same(update.callCount, 1);
-    test.same(updateSettings.callCount, 1);
-    test.done();
-  });
-};
-
```
SCdF: Since we don't want to support using these users, no tests

controllers/users.js
garethbowen: Looks right to me. Worst case we'll mark some admins as unknown. I think it's just a label in the UI.

PR #107, opened 2016-11-24
Making migrations compatible with CouchDB 2.0

SCdF:


migrations/disable-erlang-support.js
```diff
@@ -5,9 +5,9 @@ module.exports = {
   created: new Date(2016, 10, 5),
   run: function(callback) {
     db.request({
-      db: '_config',
+      db: '_node',
```
SCdF: Arguably the correct thing to do would be to just use `request` the library and don't use nano for this at all. It's a migration, so I figured the least amount of change the better.

migrations/disable-erlang-support.js
```diff
@@ -5,9 +5,9 @@ module.exports = {
...
       method: 'DELETE',
-      path: 'native_query_servers/erlang',
+      path: 'couchdb@localhost/_config/native_query_servers/erlang',
```
SCdF: I made a comment about it on the actual commit, but I'm just addressing the local node here because they are migrations so it should be fine: it's not long living code that gets run all the time.

SCdF:
> What would happen if we had a multiple node cluster? Does modifying one _node update the rest?

Almost certainly not, no. My reasoning was that they were all "legacy" migrations, and so they would only be run when you only had one node, and they all balance out to no longer be required (erlang gets enabled then disabled, we find couchdb admins and add a role so we don't have to look for couchdb admins again).

If one of those migrations was say, changing a timeout in `_config` then it should absolutely be written to cover all nodes.

garethbowen:
Ignore my comment about lucene - after talking last night I see you're only using standard mango. Merging.

PR #110, opened 2016-11-30
Move applying default settings into api. Issue: medic/medic-webapp#2951

garethbowen:


garethbowen:
Yeah fair point. I named it that so it matched config.js which matches config.js from sentinel... Maybe I'll change api/config.js to settings.js as well.

mandric:
Ah ok if the config values were only related to api or sentinel I think using `config` is fine but if we're talking about things that go to or get merged into `app_settings` we should probably use `settings` or even `app_settings` ick.

PR #114, opened 2016-12-06
Fix migrations to associate records with contacts. medic/medic-webapp…

garethbowen:
…#2970

Fix the associate-records-with-people migration to not give up so easily - we can get the contact from the clinic id if necessary. Fix the extract-person-contacts migration to store the contact.rc_code in the place_id field instead of overwriting it altogether.


mandric:
Oh, I was referring to the medic-api docs here https://github.com/medic/medic-api#optional-properties

PR #117, opened 2017-01-13
Add new message status for messages forwarded to gateway

alxndrsn:
The `sms-gateway` controller picks up all `pending` messages, and forwards these
to `medic-gateway`.  The controller then changes the messages' status to
`forwarded-to-gateway`.

Issue: #3047

### DIFF
--- a/controllers/sms-gateway.js
+++ b/controllers/sms-gateway.js
@@ -48,7 +48,7 @@ function markMessagesScheduled(messages, callback) {
   async.eachSeries(
     messages,
     function(message, callback) {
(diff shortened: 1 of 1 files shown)


alxndrsn:
I don't know where the docs are, and I wouldn't be confident to document the old state changes and what they mean.  I agree this documentation is important, but so is this fix, so if the code looks good then I would like to merge it rather than delay getting this update into the code.  A separate ticket for the documentation update might be appropriate.

mandric:
Sure I can open a separate issue, no problem.  Code looks good but I have no clue how it's supposed to work so can't review the logic.

PR #118, opened 2017-01-25
Don't change status until receipt acknowledged by gateway

garethbowen:
Also, be more tolerant of unknown statuses.

medic/medic-projects#1243

### DIFF
--- a/controllers/sms-gateway.js
+++ b/controllers/sms-gateway.js
@@ -8,9 +8,10 @@ var async = require('async'),
     recordUtils = require('./record-utils'),
     STATUS_MAP = {
       UNSENT: 'received-by-gateway',
+      PENDING: 'forwarded-by-gateway',
       SENT: 'sent',
       DELIVERED: 'delivered',
-      FAILED: 'failed'
+      FAILED: 'failed',
     };
 
 function warn() {
--- a/controllers/sms-gateway.js
+++ b/controllers/sms-gateway.js
@@ -52,20 +52,20 @@ function updateState(messageId, newState, details, callback) {
   messageUtils.updateMessage(messageId, updateBody, callback);
 }
 
-function markMessagesForwarded(messages, callback) {
-  async.eachSeries(
-    messages,
-    function(message, callback) {
-      updateState(message.id, 'forwarded-to-gateway', null, callback);
-    },
-    function(err) {
-      if (err) {
-        warn(err);
-      }
-      return callback();
-    }
-  );
-}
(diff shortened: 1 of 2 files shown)


controllers/sms-gateway.js
```diff
@@ -8,9 +8,10 @@ var async = require('async'),
     recordUtils = require('./record-utils'),
     STATUS_MAP = {
       UNSENT: 'received-by-gateway',
+      PENDING: 'forwarded-by-gateway',
```
mandric: Do we really want a `PENDING` enum that means something different than the message state `pending`?

controllers/sms-gateway.js
```diff
@@ -73,9 +80,11 @@ function getOutgoing(callback) {
         content: message.message,
       };
     });
-    markMessagesForwarded(messages, function() {
+    // TODO add this back in once this function also queries for
+    // messages in state: 'forwarded-to-gateway'
+    // markMessagesForwarded(messages, function() {
```
mandric: Can you please remind me again why we just don't implement this?  The message API currently allows for setting the state to any arbitrary string and querying for it.   The docs don't state that, but I just tried it.  https://github.com/medic/medic-api#messages

controllers/sms-gateway.js
garethbowen: "PENDING" is the gateway state and "forwarded-by-gateway" is the webapp state. PENDING is correct just in a different context. This enum is used to map between the two contexts.

controllers/sms-gateway.js
garethbowen: Gateway doesn't call the messages API. Any changes made to the gateway code are complicated because we either need to worry about backwards compatibility or try to sync updates. This is a quick fix to get 0.4 up and running, and we can do a much more robust solution in 2.10 in #3073

mandric: Gateway calls the sms-gateway API which uses the same code in message-utils.js... I'm thinking it's easy to give the gateway messages in both `pending` and `forwarded-to-gateway` states if it needs them.

garethbowen: That would require two db requests or a view change. I recommend the view change but not for 0.4.

controllers/sms-gateway.js
mandric: Can we document this?

alxndrsn: (documented in cfdd085948910e50d999bdecafa6131098a0c8a4)

PR #119, opened 2017-01-31
Message uuid migration

mandric:
Let me know if this is streamlined enough, thought I'd submit PR for initial feedback on approach.  I basically just wanted something that doesn't require `npm install` to run.  I could further streamline and make it one node script with uuid.js bundled but this was quicker and probably more flexible.  Copied/bundled the `uuid` lib with the migration.  Also added a `migrations` directory to this branch now but imagine each directory within it would just run `make` or have standalone scripts.  I could also replace the Makefile and use `set -oe` in a shell script and pipe everything together instead of using temp files.  

Does not batch, but I haven't had many problems using the _bulk_docs API with large updates. If we wanted to run updates in batches then I would move to a standalone script, but would probably cross that bridge if this setup is a problem.

I imagine this being deployed by copying the migration directory to a server then running `make`.

Issue: https://github.com/medic/medic-webapp/issues/3054


mandric:
Ok -- I just realized medic-os does not ship with make, I'll remove the temp view and look through all docs with a batch value...

PR #123, opened 2017-02-22
Don't write to finished response

garethbowen:
medic/medic-webapp#3111

### DIFF
--- a/handlers/changes.js
+++ b/handlers/changes.js
@@ -171,7 +171,8 @@ var getChanges = function(feed) {
     method: 'POST'
   }, function(err, changes) {
     if (feed.res.finished) {
-      // don't write to the response if it has already ended
+      // Don't write to the response if it has already ended. The change
+      // will be picked up in the subsequent changes request.
       return;
     }
     cleanUp(feed);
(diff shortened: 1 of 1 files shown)


handlers/changes.js
```diff
@@ -170,6 +170,11 @@ var getChanges = function(feed) {
     body: { doc_ids: _.union(feed.requestedIds, feed.validatedIds) },
     method: 'POST'
   }, function(err, changes) {
+    if (feed.res.finished) {
+      // don't write to the response if it has already ended
```
estellecomment: I'd add something like "In this case the change will be ignored, but it's ok because it'll be fetched by the next changes request that comes in"

handlers/changes.js
```diff
@@ -170,6 +170,11 @@ var getChanges = function(feed) {
...
+      return;
+    }
+    cleanUp(feed);
```
estellecomment: Is this moved up here to try avoiding reusing the same feed? 
The `abort` l. 159 looks like it would cause trouble in case of `!err`, since you haven't written to the feed yet? (don't actually know what `abort` does, maybe not)

handlers/changes.js
garethbowen: Done.

handlers/changes.js
garethbowen: Yeah this is trying to clean up earlier to win the race condition! As discussed on Hangouts this shouldn't break anything...

PR #124, opened 2017-02-27
Migration to add emit _complete to config

garethbowen:
medic/medic-webapp#1935

### DIFF
--- a/migrations/emit-complete.js
+++ b/migrations/emit-complete.js
@@ -1,0 +1,32 @@
+var path = require('path'),
+    db = require('../db');
+
+module.exports = {
+  name: 'emit-complete',
+  created: new Date(2017, 2, 27),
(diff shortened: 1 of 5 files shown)


migrations/emit-complete.js
```diff
@@ -0,0 +1,32 @@
...
+        // no rules configured
+        return callback();
+      }
+
+      if (rules.indexOf('_complete') !== -1) {
+        // complete task already configured
+        return callback();
+      }
+
+      rules = rules.replace(/(.*)\}[ \n\t]*\}[ \n\t]*$/, '$1 emit("_complete", { _id: true });}}');
```
SCdF: Regex of doom. Can we get a comment with examples of what this catches, and what it gets turned into with the replace?

(will still merge, do it in master).

PR #125, opened 2017-02-28
Allow accessing SMS api endpoint with trailing slash

alxndrsn:
Issue: medic/medic-webapp#3152

I think this is helpful for tech leads when testing gateway setup.  I'm not 100% convinced having the API served from either endpoint is correct though - perhaps it should be `301`ing from `../sms/` to `../sms`?

### DIFF
--- a/server.js
+++ b/server.js
@@ -191,33 +191,39 @@ app.get('/api/upcoming-due-dates', function(req, res) {
   handleAnalyticsCall(req, res, upcomingDueDates);
 });
 
-app.get('/api/sms', function(req, res) {
-  auth.check(req, 'can_access_gateway_api', null, function(err) {
-    if (err) {
-      return serverUtils.error(err, req, res);
-    }
-    smsGateway.get(function(err, obj) {
+(function() {
+  var handleSms_GET = function(req, res) {
+    auth.check(req, 'can_access_gateway_api', null, function(err) {
       if (err) {
-        return serverUtils.error(err, res);
+        return serverUtils.error(err, req, res);
       }
-      res.json(obj);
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -191,33 +191,39 @@ app.get('/api/upcoming-due-dates', function(req, res) {
...
 });
 
-app.get('/api/sms', function(req, res) {
-  auth.check(req, 'can_access_gateway_api', null, function(err) {
-    if (err) {
-      return serverUtils.error(err, req, res);
-    }
-    smsGateway.get(function(err, obj) {
+(function() {
+  var handleSms_GET = function(req, res) {
```
SCdF: what's up with the underscores? Wouldn't the naming convention / "standard" make these functions `handleSmsGet` and `handleSmsPost`?

server.js
```diff
@@ -191,33 +191,39 @@ app.get('/api/upcoming-due-dates', function(req, res) {
...
+        }
+        res.json(obj);
+      });
     });
-  });
-});
+  };
+  app.post('/api/sms', jsonParser, handleSms_POST);
+  app.post('/api/sms/', jsonParser, handleSms_POST);
+}());
```
SCdF: what problem does wrapping this code in an executing function solve? It makes this new code (along with the function naming convention) rather foreign in this file. I would prefer that we tried to keep all the looking the same and using the same paradigms. I don't massively care _which_ paradigm we use, just as long as we don't introduce more than we need.

server.js
alxndrsn: Dunno, just seemed clearer to me.

PR #126, opened 2017-03-02
3058 manual migrations

SCdF:
Ideally don't squash this, each commit makes sense on its own, and distinct from each other.


migrations/extract-data-record-content.js
```diff
@@ -0,0 +1,102 @@
...
+            docs = docs.map(attachmentIse).filter(function(i) {
+              return i;
+            });
+
+            if (docs.length === 0) {
+              // The view we're using gets all data records, not just XML ones
+              // so there is a good chance entire batches will be filtered out
+              // TODO: when we upgrade to CouchDB2.0 this is a great place to use
+              //       mango filters to target XML forms better
+              return callback();
```
garethbowen: Use `return` or `else`. Both looks weird.

migrations/extract-data-record-content.js
```diff
@@ -0,0 +1,102 @@
...
+var incrementPercentTarget = function(percentTarget) {
+  return percentTarget + PERCENT_REPORT_CHUNKS;
+};
+
+module.exports = {
+  name: 'extract-data-record-content',
+  created: new Date(2017, 0, 6),
+  run: function(callback) {
+    console.log('Finding data_records...');
+    db.medic.view('medic', 'data_records', function(err, body) {
```
garethbowen: I think fetching _all_ data record stubs will cause issues on large DBs (eg: LG). Because of this, other migrations use limit and skip to batch the view request and include docs at the same time. You get `total_rows` in the view response so you can still work out percentage complete (which is nice btw!). You'd have to change your logic regarding done a little, but overall it'd be more reliable.

migrations/extract-data-record-content.js
```diff
@@ -0,0 +1,102 @@
...
+};
+
+
+var incrementPercentTarget = function(percentTarget) {
+  return percentTarget + PERCENT_REPORT_CHUNKS;
+};
+
+module.exports = {
+  name: 'extract-data-record-content',
+  created: new Date(2017, 0, 6),
```
garethbowen: Set this to today's date, unless you have good reason?

migrations/extract-data-record-content.js
```diff
@@ -0,0 +1,102 @@
+var _ = require('underscore'),
```
garethbowen: This is still in the `migrations` directory? What's going to stop this from running automatically? Or do we want it to run automatically now?

scripts/run-migration-manually.js
```diff
@@ -0,0 +1,32 @@
+#!/bin/node
+
+// Intentionally not ES6 because it may be run on old systems with an old node
+
+var scriptLocation = process.argv[2];
+
+if (!scriptLocation) {
+  console.error('You must provide the script to run as argument.');
```
garethbowen: Change "script" to "migration" to make it clear this only runs migrations.

garethbowen:
Ok... I made comments on the individuals commits which is a little confusing in this UI but I trust you can find them. I'm a little uncomfortable having a scripts directory in each repo because it might make it hard to find things. Currently the api scripts all seem to be related to testing. How would you feel about putting the migration runner in `medic-webapp/scripts` so it's easier to find?

migrations/extract-data-record-content.js
SCdF: That was the day it was originally written. I presume that's the same it's supposed to be right?

migrations/extract-data-record-content.js
SCdF: Hmm good point. I had figured that just getting the list of ids was actually fine, but I suppose with larger DBs this might not be the same. I'll rejig it to have two layers of batching.

NB: this with percentages only works in this case because we're not filtering the view at all. In other situations we might need to batch load the view in chunks but still upfront to get percentages (somewhat ironically a % is only important when there is a lot of data, which in turn means loading ids upfront is a pita)

migrations/extract-data-record-content.js
SCdF: Yeah ugh. I have no idea how to make JS look nice, all the paradigms look wrong to me :-(

I'll drop the else.

_(FYI I did this because I instinctively put `return` before a call to `callback`, because I've been stung by that stupid bit of cruft code too many times. Can't wait for promises.)_

SCdF:
@garethbowen yep I can move the script to the medic-webapp directory.

SCdF:
@garethbowen I moved the script to medic-webapp directly in master. I have made reasonably substantial changes to the script to support view batching, please take another look

SCdF:
(oh, it should be squashed now)

SCdF:
Yeah, IDK, it was how my brain solved that problem. I cannot think of any reason why my way would be better than what you suggest, and your version would be simpler. I'm going to leave it though, because as you say, it may never run.

SCdF:
Woah apparently you can't close and *then* merge. TIL

PR #129, opened 2017-03-07
Support both couchdb1&2 medic/medic-webapp#3187

SCdF:


SCdF:
Whoops, I meant to flip those to be consistent. I'll do that then merge.

mandric:
Hey could we have a call on this?  I'm curious to know what went into this (lazy to read through all the code and issues).

SCdF:
@mandric sure ping me when I'm around. The code is actually really minimal, I'm happy to walk through it.

PR #130, opened 2017-03-23
3226 env var check

SCdF:
Checks to make sure you've set the right env vars, and that you're on the right version of Node.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,12 @@ Export a `COUCH_URL` env variable so sentinel knows what database to use. e.g.
 export COUCH_URL='http://admin:pass@localhost:5984/medic'
 ```
 
+If you are using CouchDB2.0 you need to also provide your node name. e.g.
+
+```
+export COUCH_NODE_NAME=couchdb@localhost node server.js
+```
+
 If you're using couchdb-lucene in any configuration other than the standard (on the same server as CouchDb on port 5985) you can specify it with the `LUCENE_URL` env variable:
 
 ```
--- a/server.js
+++ b/server.js
@@ -293,7 +293,7 @@ app.all([
 
       res
         .set('Content-Type', format.contentType)
-        .set(`Content-Disposition', 'attachment; filename=${filename}`);
+        .set('Content-Disposition', `attachment; filename=${filename}`);
 
       if (_.isFunction(exportDataResult)) {
         // wants to stream the result back
@@ -613,7 +613,8 @@ app.put(auditPath, audit);
 app.post(auditPath, audit);
 app.delete(auditPath, audit);
 
-app.all('*', proxy.web);
+app.all('*', (req, res) =>
+  proxy.web(req, res));
 
 proxy.on('error', (err, req, res) => serverUtils.serverError(JSON.stringify(err), req, res));
 
--- a/server.js
+++ b/server.js
@@ -653,17 +653,17 @@ proxyForAuditing.on('error', function(err, req, res) {
 
 const nodeVersionCheck = callback => {
   try {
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -73,6 +73,12 @@ Export a `COUCH_URL` env variable so sentinel knows what database to use. e.g.
 export COUCH_URL='http://admin:pass@localhost:5984/medic'
 ```
 
+If you are using CouchDB2.0 you need to also provide your node name. e.g.
+
+```
+export COUCH_NODE_NAME=couchdb@localhost node server.js
```
alxndrsn: I assume this is the default node name; please confirm!

server.js
```diff
@@ -651,20 +651,59 @@ proxyForAuditing.on('error', function(err, req, res) {
...
 });
 
-var couchDbVersionCheck = function(callback) {
-  db.getCouchDbVersion(function(err, version) {
-    console.log('CouchDB Version:', version);
+const nodeVersionCheck = callback => {
+  try {
+    console.log('Node Version:', process.version);
+
+    const version = process.versions.node.match(/(\d)+\.(\d)+\.(\d)+/)[1];
```
alxndrsn: I probably don't want to know why you have to reference `process.version` on one line, and then this horrible regex against a different string on the next.

server.js
```diff
@@ -651,20 +651,59 @@ proxyForAuditing.on('error', function(err, req, res) {
...
+
+    if (Number(version[1] <= 4)) {
```
alxndrsn: Might be nice to be consistent and say `< 5` here, as per the comment on the following line.  Also, the code below also uses `<` instead of `<=`.

server.js
alxndrsn: Actually this `[1]` seems a bit dodgy - IIUC this will give you the major version number.  But below, you're trying to get sub-parts of `version`, which will give specific characters from the string matched on this line.

E.g.
```
> process.versions.node
'1.2.3'
> majorVersion = process.versions.node.match(/(\d)+\.(\d)+\.(\d)+/)[1];
'1'
> majorVersion[0]
'1'
> majorVersion[1]
undefined
```

I assume that `const version` is meant to be assigned to the full version string (`1.2.3`).  If so, it might be nice to use it in the debug output above as well.  If not, it might be better called `majorVersion`.

server.js
```diff
@@ -651,20 +651,59 @@ proxyForAuditing.on('error', function(err, req, res) {
...
+
+  if (failures.length) {
+    callback('At least one required environment variable was not set:\n' + failures.join('\n'));
+  } else {
+    callback();
+  }
 };
 
+const couchDbVersionCheck = callback =>
+  db.getCouchDbVersion((err, version) => {
```
alxndrsn: What happens if `err` is set?

README.md
SCdF: It's the default that I have. I didn't change it to that. I presume this is what you have as well? If not we can change it to whatever

server.js
SCdF: `process.version` is the "display" version, `process.versions.node` is the non display version. The only difference AFAICT is that one has a "v" in front of it. We should probably just use the latter.

SCdF: Hey you're right, that is dodgy. I have no idea why I did it that way! (It's also done that way in Sentinel, I'll change Sentinel's to whatever I do here as well)

server.js
SCdF: Done

server.js
SCdF: Good call, I've passed it through the async.series so it should fail properly

PR #131, opened 2017-03-30
The national_admin role is now sufficient to edit users

SCdF:
Issue: medic/medic-webapp#3262

### DIFF
--- a/migrations/add-national_admin-role.js
+++ b/migrations/add-national_admin-role.js
@@ -1,0 +1,29 @@
+const db = require('../db');
+
+module.exports = {
+  name: 'add-national_admin-role',
+  created: new Date(2017, 3, 30),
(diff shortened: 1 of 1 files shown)


migrations/add-national_admin-role.js
```diff
@@ -0,0 +1,34 @@
+const _ = require('underscore'),
+      db = require('../db'),
+      async = require('async');
```
garethbowen: Consider using: `const series = require('async/series');`

SCdF: OK I'll do this because I haven't used it yet… but in my mind I don't actually like this feature. If it helps code be smaller or something that's one thing, but if it's just because people don't want to constantly specify the namespace… I *like* specifying the namespace!

PR #132, opened 2017-03-30
Only redirect to login page if requesting client is human

alxndrsn:
This means that android apps like medic-collect and medic-gateway will receive
status code 401 instead of 302 when they fail to authenticate.

Issue: #3118

...because we all like parsing useragents.

### DIFF
--- a/client_is_human.js
+++ b/client_is_human.js
@@ -1,4 +1,4 @@
-var DALVIK_DETECT = new RegExp('^Dalvik/');
+const DALVIK_DETECT = new RegExp('^Dalvik/');
 
 /**
  * Check an incoming request to see if it's from a human browser, or from an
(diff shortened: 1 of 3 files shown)


client_is_human.js
```diff
@@ -0,0 +1,23 @@
+var DALVIK_DETECT = new RegExp('^Dalvik/');
```
SCdF: You can start using ES6 now in api.

If you don't want to / would rather stick with the old style, we should have a conversation about it as a team, because if we're not all pulling in the same direction our code is going to get pretty weird pretty quickly. I'd rather not have half of writing ES6 and the other half not.

client_is_human.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+module.exports = function(req) {
+  var ua = req.headers && req.headers['user-agent'];
+
+  if (!ua) {
+    return false;
+  }
+
+  if (DALVIK_DETECT.test(ua)) {
+    return ua.indexOf('medic.gateway') === -1;
```
SCdF: FWIW you can use `.includes` now, i.e. `return !us.includes('medic-gateway')`

client_is_human.js
alxndrsn: i.e. `const` instead of `var`?  or something more than that?

PR #134, opened 2017-03-31
Missed a bit!

alxndrsn:
Suggested squashed commit message below

-----

Prompt for basic auth on authentication failure if the client is non-human

medic/medic-webapp#3118


alxndrsn:
Fixing,,,

alxndrsn:
@SCdF Some further review, please.  Not sure how I missed this second place to do redirects to the login page.

I'm not entirely happy with the name or flow of `notLoggedIn()`, or the variable `showPrompt`: it looks a little bit arcane to me - `showPrompt` seems to have been added later, and actually there are (arguably) two different kinds of prompt that might be displayed - the browser's built-in basic auth prompt vs the app's login screen.  I'm surprised we can't return the `WWW-Authenticate` header when writing a JSON response too, but perhaps best not to fiddle...

Please pass on to gareth if you're not happy reviewing (he'd probably want to see https://github.com/medic/medic-api/commit/fec0e015b0484b7be4b58607d1b4e88c7e00c58c as well).

alxndrsn:
@garethbowen Please review, including the already-merged commit at https://github.com/medic/medic-api/commit/fec0e015b0484b7be4b58607d1b4e88c7e00c58c.

Also I'm starting to think this would benefit from end-to-end testing, but wouldn't necessarily want that to hold up this PR.  What do you think?

garethbowen:
@alxndrsn One comment added to the commit: https://github.com/medic/medic-api/commit/fec0e015b0484b7be4b58607d1b4e88c7e00c58c

Also, can you rename the `client_is_human` module to `client-is-human` (kebab style) to match the other files?

Otherwise, all good!

alxndrsn:
> Also, can you rename the client_is_human module to client-is-human (kebab style) to match the other files?

@garethbowen inteteresting!  I was considering renaming it to `is_client_human`, so now i have an opportunity.  Do you have any preference for either of:
- [ ] `is-client-human`
- [ ] `client-is-human`
?

garethbowen:
@alxndrsn I prefer the question than the statement, ie: is-client-human

PR #135, opened 2017-03-31
Redirect `/api/sms/` requests to `/api/sms`

alxndrsn:
Issue: medic/medic-webapp#3152

### DIFF
--- a/server.js
+++ b/server.js
@@ -193,6 +193,9 @@ app.get('/api/upcoming-due-dates', function(req, res) {
   handleAnalyticsCall(req, res, upcomingDueDates);
 });
 
+app.get('/api/sms/', function(req, res) {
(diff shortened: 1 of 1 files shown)


alxndrsn:
> Is there a reason why this endpoint is different?

Not a principled one.  But that's how it is... I'm definitely open to suggestions if you can think of a better alternative; if that means rigorously testing gateway to see if it properly follows redirects, and fixing future versions & documentation to include the trailing slash, then very happy to do that.

garethbowen:
If it's hardcoded on the gateway side, just go with this as is.

alxndrsn:
Sure?  Last chance to change...!





Actually I guess we could switch it round later with the same challenge, if we felt like.

PR #136, opened 2017-04-02
Migrate migration log to have predictable id

garethbowen:
Make it easier to find the migration log as it turns out it's useful
information for humans as well as machines.

medic/medic-webapp#3294

### DIFF
--- a/migrations.js
+++ b/migrations.js
@@ -3,33 +3,75 @@ var fs = require('fs'),
     async = require('async'),
     db = require('./db');
 
-var hasRun = function(meta, migration) {
-  if (!meta || !meta.migrations) {
+var MIGRATION_LOG_ID = 'migration-log',
+    MIGRATION_LOG_TYPE = 'meta';
+
+var hasRun = function(log, migration) {
+  if (!log || !log.migrations) {
     return false;
   }
-  return meta.migrations.indexOf(migration.name) !== -1;
+  return log.migrations.indexOf(migration.name) !== -1;
 };
 
 var error = function(migration, err) {
   return 'Migration "' + migration.name + '" failed with: ' + JSON.stringify(err);
 };
 
-var getMeta = function(callback) {
-  db.medic.view('medic-client', 'doc_by_type', { include_docs: true, key: [ 'meta' ] }, function(err, meta) {
+var getLogWithView = function(callback) {
+  var options = { include_docs: true, key: [ MIGRATION_LOG_TYPE ] };
+  db.medic.view('medic-client', 'doc_by_type', options, function(err, result) {
     if (err) {
       return callback(
         new Error(
           'Could not run migrations without doc_by_type view. Update ddoc. ' +
           JSON.stringify(err)));
     }
-    meta = meta && meta.rows && meta.rows[0] && meta.rows[0].doc;
-    if (!meta) {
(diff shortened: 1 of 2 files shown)


alxndrsn:
I don't see an accompanying commit in `medic-webapp`; is this doc only used in `medic-api`?

garethbowen:
@alxndrsn Yeah it's only used for recording which migrations have been executed.

migrations.js
```diff
@@ -3,33 +3,75 @@ var fs = require('fs'),
...
+
+var deleteOldLog = function(oldLog, callback) {
+  if (!oldLog) {
+    return callback();
+  }
+  oldLog._deleted = true;
+  db.medic.insert(oldLog, callback);
+};
+
+var createLog = function(callback) {
```
alxndrsn: Not convinced by this name - this is the function that migrates from old log doc to the new one IIUC, so perhaps `migrateLogToKnownId()` or something would be more appropriate?

migrations.js
```diff
@@ -79,15 +121,18 @@ var runMigrations = function(meta, migrations, callback) {
 
 module.exports = {
   run: function(callback) {
-    getMeta(function(err, meta) {
+    module.exports.get(function(err, migrations) {
```
alxndrsn: Appreciate this is not part of this change, but is `get()` only exported for unit tests?

migrations.js
```diff
@@ -79,15 +121,18 @@ var runMigrations = function(meta, migrations, callback) {
...
       if (err) {
         return callback(err);
       }
-      module.exports.get(function(err, migrations) {
+      if (!migrations.length) {
```
alxndrsn: Is this an alternative to checking `if(err)`?  Can't see why the migrations list would ever be falsy, otherwise.

migrations.js
```diff
@@ -3,33 +3,75 @@ var fs = require('fs'),
...
+
+var getLog = function(callback) {
+  db.medic.get(MIGRATION_LOG_ID, function(err, doc) {
+    if (!err) {
+      return callback(null, doc);
     }
-    if (!meta.migrations) {
-      meta.migrations = [];
+    if (err.statusCode === 404) {
+      return createLog(callback);
```
alxndrsn: Doesn't look like `createLog()` would ever return anything.  Is this return just to shortcut out of this function, or is there meant to be a chained promise here?

migrations.js
garethbowen: It migrates the old log if it's there (upgrade), or creates an empty one if it's not there (fresh install). I've renamed it to `createMigrationLog` because it's creating the new doc style ont way or another.

migrations.js
garethbowen: There are no promises in this code anywhere, mostly because we still use nano which is callback based. One day...

I've cleaned up the code a little to hopefully make it easier to read.

migrations.js
garethbowen: Yeah I think that's right. IIRC when writing this originally it was too hard to mock out reading the migrations dir.

migrations.js
garethbowen: That's a good point. We have hardcoded at least one migration! I got a bit overzealous with edge cases. I'll remove it.

PR #140, opened 2017-04-11
Send basic auth header to Collect

alxndrsn:
Issue: medic/medic-webapp#3118

Other commits made in attempting to resolve this issue:
* fec0e015b0484b7be4b58607d1b4e88c7e00c58c
* 819c24eed00b017c6639a4d16a898c6b7bb5d205

It's not currently clear to me why the `WWW-Authenticate` header is set in so many places.

### DIFF
--- a/server.js
+++ b/server.js
@@ -566,7 +566,7 @@ app.post(pathPrefix + '_changes', jsonParser, changesHander);
 var writeHeaders = function(req, res, headers, redirectHumans) {
   res.oldWriteHead = res.writeHead;
   res.writeHead = function(_statusCode, _headers) {
-    // hardcode this so we never show the basic auth prompt
+    // hardcode this so we do not show the basic auth prompt by default
     res.setHeader('WWW-Authenticate', 'Cookie');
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -566,20 +566,26 @@ app.post(pathPrefix + '_changes', jsonParser, changesHander);
 var writeHeaders = function(req, res, headers, redirectHumans) {
   res.oldWriteHead = res.writeHead;
   res.writeHead = function(_statusCode, _headers) {
-    // hardcode this so we never show the basic auth prompt
+    // hardcode this so we do not show the basic auth prompt by default
     res.setHeader('WWW-Authenticate', 'Cookie');
```
garethbowen: Can we move this line down into the `isClientHuman` block so we're not writing the header twice?

alxndrsn: It could be a change of behaviour, as:
* the header may be required for non-`401` responses
* that header can also be overwritten in the `headers.forEach(...)` loop below

So, I'm not sure.

PR #143, opened 2017-04-13
3371 unbreak api

estellecomment:
medic/medic-webapp#3371

### DIFF
--- a/tests/unit/migrations/namespace-form-fields.js
+++ b/tests/unit/migrations/namespace-form-fields.js
@@ -5,16 +5,12 @@ var sinon = require('sinon'),
     migration = require('../../../migrations/namespace-form-fields');
 
 exports.tearDown = function (callback) {
-  console.log('-----restore');
-  console.log(db.medic.bulk.isSinonProxy);
   utils.restore(
     db.medic.view,
     db.medic.bulk,
     config.get,
     config.load
   );
-  console.log('-----restoreddd');
-  console.log(db.medic.bulk.isSinonProxy);
   callback();
 };
 
@@ -68,7 +64,6 @@ exports['run does nothing if no data records'] = function(test) {
   sinon.stub(config, 'get').returns({});
   sinon.stub(db.medic, 'bulk').callsArgWith(1, null, null);
   migration.run(function(err) {
-    console.log('whyyyyy');
     test.equals(err, undefined);
     test.equals(getView.callCount, 1);
     test.done();
--- a/tests/unit/migrations/namespace-form-fields.js
+++ b/tests/unit/migrations/namespace-form-fields.js
@@ -5,12 +5,16 @@ var sinon = require('sinon'),
     migration = require('../../../migrations/namespace-form-fields');
 
 exports.tearDown = function (callback) {
+  console.log('-----restore');
+  console.log(db.medic.bulk.isSinonProxy);
   utils.restore(
     db.medic.view,
     db.medic.bulk,
     config.get,
     config.load
   );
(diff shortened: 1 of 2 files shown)


tests/unit/migrations/namespace-form-fields.js
```diff
@@ -4,6 +4,39 @@ var sinon = require('sinon'),
     utils = require('../utils'),
     migration = require('../../../migrations/namespace-form-fields');
 
+// Arguments : docArrayForFirstBatch, docArrayForSecondBatch, ...
+var makeStubs = function() {
```
SCdF: This looks like a job for ES6!

You want varargs right?

```js
const makeStubs = (...batches) => {
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator

tests/unit/migrations/namespace-form-fields.js
```diff
@@ -4,6 +4,39 @@ var sinon = require('sinon'),
...
+  if (arguments.length === 0) {
+    getView.callsArgWith(3, null,
+      {
+        total_rows: 0,
+        rows: []
+      });
+  } else {
+    var viewBatches = [];
+    var totalRows = 0;
+    for (var i = 0; i < arguments.length; i++) {
```
SCdF: Fwiw you can use `arguments.forEach(argument => {...` here instead of a for loop and referencing by index.

SCdF: Actually looking at this whole section I think it's basically a reduce?:

```js
const [viewBatches, totalRows] = arguments.reduce((acc, argument) => {
  acc[0].push(argument);
  acc[1] += argument.length;
}, [[], 0]);
```

estellecomment: I knooooooooow
But arguments isn't actually an array, it's this weird object. So the reduce breaks. Very disappointing.

tests/unit/migrations/namespace-form-fields.js
```diff
@@ -4,6 +4,37 @@ var sinon = require('sinon'),
     utils = require('../utils'),
     migration = require('../../../migrations/namespace-form-fields');
 
+// Arguments : docArrayForFirstBatch, docArrayForSecondBatch, ...
+var makeStubs = (...viewBatches) => {
+  console.log(viewBatches);
```
SCdF: Probably not intentional?

tests/unit/migrations/namespace-form-fields.js
```diff
@@ -4,6 +4,37 @@ var sinon = require('sinon'),
     utils = require('../utils'),
     migration = require('../../../migrations/namespace-form-fields');
 
+// Arguments : docArrayForFirstBatch, docArrayForSecondBatch, ...
```
SCdF: You don't need this comment anymore right? You can see what to pass because you have a parameter

tests/unit/migrations/namespace-form-fields.js
```diff
@@ -4,6 +4,37 @@ var sinon = require('sinon'),
...
+  console.log(viewBatches);
+  var getView = sinon.stub(db.medic, 'view');
+  if (viewBatches.length === 0) {
+    getView.callsArgWith(3, null,
+      {
+        total_rows: 0,
+        rows: []
+      });
+  } else {
+    var totalRows = viewBatches.reduce(function(total, batch) {
```
SCdF: FWIW ES6: this can just be `viewBatches.reduce((total, batch) => total + batch.length, 0);`

PR #144, opened 2017-04-21
Using sinon.sandbox for clearer, safer tests

SCdF:
Replacing our pattern of passing a list of things that have maybe been
stubbed to a utility that restores them, and replacing it with sinon
sandboxes, the correct way to deal with this kind of thing.

This is better because it's fool-proof, you do not have to know what
things may or may not have been stubbed.

See: http://sinonjs.org/releases/v2.1.0/sandbox/sinon

### DIFF
--- a/tests/unit/auth.js
+++ b/tests/unit/auth.js
@@ -1,17 +1,12 @@
-var request = require('request'),
-    url = require('url'),
-    sinon = require('sinon'),
-    auth = require('../../auth'),
-    utils = require('./utils'),
-    config = require('../../config'),
-    db = require('../../db');
-
-exports.setUp = function(callback) {
-  callback();
-};
+const request = require('request'),
+      url = require('url'),
+      sinon = require('sinon').sandbox.create(),
+      auth = require('../../auth'),
+      config = require('../../config'),
+      db = require('../../db');
 
 exports.tearDown = function (callback) {
-  utils.restore(request.get, request.head, url.format, config.get);
+  sinon.restore();
   callback();
 };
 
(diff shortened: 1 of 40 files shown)


tests/unit/auth.js
```diff
@@ -1,17 +1,12 @@
...
-    sinon = require('sinon'),
-    auth = require('../../auth'),
-    utils = require('./utils'),
-    config = require('../../config'),
-    db = require('../../db');
-
-exports.setUp = function(callback) {
-  callback();
-};
+const request = require('request'),
```
alxndrsn: i see what you're up to here...

tests/unit/controllers/missed-appointments.js
```diff
@@ -51,23 +50,23 @@ exports['get returns zero if all registrations have delivered'] = function(test)
   var fti = sinon.stub(db, 'fti');
   fti.onFirstCall().callsArgWith(2, null, {
     rows: [
-      { 
-        doc: {
```
alxndrsn: this lot doesn't seem to be relevant.  if you really want to make this change, just push to `master` IMO

tests/unit/controllers/monthly-deliveries.js
```diff
@@ -117,7 +116,7 @@ exports['get returns monthly deliveries count'] = function(test) {
       { month: 'Sep 1969', count: 1 }, // delivery report for patient 4
       { month: 'Oct 1969', count: 0 },
       { month: 'Nov 1969', count: 0 },
-      { month: 'Dec 1969', count: 0 } 
+      { month: 'Dec 1969', count: 0 }
```
alxndrsn: Missing a `,` 😛

tests/unit/migrations/associate-records-with-people.js
```diff
@@ -192,7 +191,7 @@ var outgoingMessageToPerson = {
                   phone: '+2884615402'
                 },
                 parent: {
-                  
+
```
alxndrsn: Why is there an empty parent?  Can't you just leave the whole thing out?

PR #145, opened 2017-04-22
3073 message delivery statuses

estellecomment:
medic/medic-webapp#3073

### DIFF
--- a/controllers/sms-gateway.js
+++ b/controllers/sms-gateway.js
@@ -62,6 +62,8 @@ function markMessagesForwarded(messages, callback) {
     },
     function(err) {
       if (err) {
+        // No error throwing here, because no big deal : 'forwarded-to-gateway' status
+        // wasn't saved to DB, so message will stay in 'pending' status and be retried next time.
         warn(err);
       }
       return callback();
--- a/controllers/sms-gateway.js
+++ b/controllers/sms-gateway.js
@@ -54,23 +54,24 @@ function updateState(messageId, newState, details, callback) {
   messageUtils.updateMessage(messageId, updateBody, callback);
 }
 
-// function markMessagesForwarded(messages, callback) {
-//   async.eachSeries(
-//     messages,
-//     function(message, callback) {
-//       updateState(message.id, 'forwarded-to-gateway', null, callback);
-//     },
-//     function(err) {
-//       if (err) {
-//         warn(err);
-//       }
-//       return callback();
-//     }
-//   );
-// }
+function markMessagesForwarded(messages, callback) {
+  async.eachSeries(
+    messages,
+    function(message, callback) {
+      updateState(message.id, 'forwarded-to-gateway', null, callback);
+    },
+    function(err) {
+      if (err) {
+        warn(err);
+      }
+      return callback();
+    }
+  );
+}
 
 function getOutgoing(callback) {
-  messageUtils.getMessages({ state: 'pending' }, function(err, pendingMessages) {
+  messageUtils.getMessages({ states: ['pending', 'forwarded-to-gateway'] },
+    function(err, pendingMessages) {
     if (err) {
       warn(err);
       return callback(null, []);
@@ -82,11 +83,9 @@ function getOutgoing(callback) {
         content: message.message,
       };
     });
-    // TODO add this back in once this function also queries for
-    // messages in state: 'forwarded-to-gateway'
-    // markMessagesForwarded(messages, function() {
+    markMessagesForwarded(messages, function() {
       callback(null, messages);
-    // });
+    });
   });
 }
 
--- a/tests/e2e/controllers/messages.js
+++ b/tests/e2e/controllers/messages.js
@@ -1,0 +1,76 @@
+const assert = require('chai').assert,
+  PouchDB = require('pouchdb'),
+  request = require('request-promise-native'),
+  Url = require('url'),
+  utils = require('../utils');
+
+const adminDb = new PouchDB(process.env.COUCH_URL);
+
+const data_record = {
+  _id: 'my_data_record',
+  errors: [],
+  form: null,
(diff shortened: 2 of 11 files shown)


controllers/sms-gateway.js
```diff
@@ -54,23 +54,24 @@ function updateState(messageId, newState, details, callback) {
...
-// }
+function markMessagesForwarded(messages, callback) {
+  async.eachSeries(
+    messages,
+    function(message, callback) {
+      updateState(message.id, 'forwarded-to-gateway', null, callback);
+    },
+    function(err) {
+      if (err) {
+        warn(err);
```
garethbowen: It might be worth a comment here to explain why you're ok with just warning on `err`.

tests/e2e/controllers/messages.js
```diff
@@ -0,0 +1,208 @@
+const assert = require('chai').assert,
+  PouchDB = require('pouchdb'),
+  request = require('request-promise-native'),
+  Url = require('url'),
```
garethbowen: IMO use lowercase for `Url` and `PouchDB` constants. You could argue for camelcase for `pouchDB` or `pouchDb` but I think lower is most obvious.

tests/e2e/controllers/messages.js
```diff
@@ -0,0 +1,208 @@
...
+
+const data_record = {
+  _id: 'my_data_record',
+  errors: [],
+  form: null,
+  from: '0211111111',
+  reported_date: 1432801258088,
+  tasks: [
+   {
+       state: 'sent',
```
garethbowen: Nitpick: use 2 spaces for indentation not 4. Depending on your tool there is probably an easy way to convert this.

tests/e2e/controllers/messages.js
```diff
@@ -0,0 +1,208 @@
...
+               to: '0222222222',
+               uuid: pendingId
+           }
+       ]
+   }
+  ]
+};
+
+const apiRequest = (endpoint, queryParams) => {
+  var url = Url.parse(process.env.API_URL);
```
garethbowen: Is this right? Most e2e tests use the constants.js to get the api URL. Can you use this too?

tests/e2e/controllers/messages.js
```diff
@@ -0,0 +1,208 @@
...
+               message: 'A vulnerable child has been referred to you. (High Priority)',
+               to: '0222222222',
+               uuid: pendingId
+           }
+       ]
+   }
+  ]
+};
+
+const apiRequest = (endpoint, queryParams) => {
```
garethbowen: Isn't this duplicated by utils.request? This is nicer because it uses the library to achieve (I think) the same thing. Maybe improve the utils.request function to use this library too?

tests/e2e/controllers/messages.js
```diff
@@ -0,0 +1,208 @@
...
+      .then(() => adminDb.put(data_record))
+      .then((result) => {
+        data_record_rev = result.rev;
+        done();
+      })
+      .catch(done);
+  });
+
+  afterEach(function(done) {
+    adminDb.remove(data_record._id, data_record_rev)
```
garethbowen: Why do you supply the `rev`? Are you trying to catch document update conflicts?

tests/e2e/controllers/messages.js
```diff
@@ -0,0 +1,208 @@
...
+
+  it('should fetch all messages', function(done) {
+    apiRequest('/api/v1/messages')
+    .then((result) => {
+      // TODO stop emitting everything twice : https://github.com/medic/medic-webapp/issues/3400
+      // assert.equal(result.length, 3);
+      assert.equal(result.length, 6);
+      done();
+    })
+    .catch((err) => done(err));
```
garethbowen: Why not `.catch(done)`? Good ol' ES0  :)

tests/e2e/controllers/messages.js
```diff
@@ -0,0 +1,208 @@
+const assert = require('chai').assert,
+  PouchDB = require('pouchdb'),
+  request = require('request-promise-native'),
+  Url = require('url'),
+  utils = require('../utils');
+
+const adminDb = new PouchDB(process.env.COUCH_URL);
```
garethbowen: Use constants.js for this.

tests/e2e/controllers/messages.js
```diff
@@ -0,0 +1,208 @@
...
+      assert.ok(
+        arr[i-1].sending_due_date >=
+        arr[i].sending_due_date,
+        'Expected ' + arr[i-1].sending_due_date + ' >= ' + arr[i].sending_due_date);
+    }
+  };
+
+  var assertSortedByAscendingDate = (arr) => {
+    for (let i = 1; i < arr.length; i++) {
+      assert.ok(
```
garethbowen: You can use [assert.isBelow](http://chaijs.com/api/assert/#method_isbelow) here which might print you a pretty error message for free!

tests/e2e/controllers/messages.js
estellecomment: `PouchDB` is used all over the codebase, so I'm leaving it. Changed the other.

tests/e2e/controllers/messages.js
estellecomment: two different apis for deleting : either provide the whole doc which includes the rev, or provide just `id` and `rev` which is what I'm using here.

tests/e2e/controllers/messages.js
estellecomment: Arrows!!! @SCdF said arrows!!!

tests/e2e/controllers/messages.js
estellecomment: I'd need `isBelowOrEqual`, which doesn't exist... :(

PR #146, opened 2017-04-24
Add delete_fixtures for api e2e tests. medic/medic-webapp#3409

estellecomment:
medic/medic-webapp#3409


scripts/e2e/delete_fixtures
```diff
@@ -0,0 +1,35 @@
...
+	fi
+}
+
+delete_user() {
+	local username="$1"
+	echo "$API_URL/api/v1/users/$1"
+	curl -X DELETE "$API_URL/api/v1/users/$1"
+}
+
+echo "[$0] deleting users created by test fixtures..."
```
alxndrsn: Why only 1 `echo` here?  (Not that I can remember why there were 2 elsewhere...)

scripts/e2e/delete_fixtures
```diff
@@ -0,0 +1,35 @@
...
+	local docUrl="$COUCH_URL/$1"
+	local rev="$(curl $docUrl | jq --raw-output ._rev)"
+	if [[ "$rev" != "null" ]]; then
+		curl -X DELETE "$docUrl?rev=$rev"
+	fi
+}
+
+delete_user() {
+	local username="$1"
+	echo "$API_URL/api/v1/users/$1"
```
alxndrsn: This looks like debug; should it be removed?  If not, some context would be helpful

scripts/e2e/delete_fixtures
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash -eu
+
+# Does the inverse of create_fixtures
+echo && echo "[$0] deleting fixtures for api e2e tests..."
+
+delete_doc() {
+	local docUrl="$COUCH_URL/$1"
+	local rev="$(curl $docUrl | jq --raw-output ._rev)"
```
alxndrsn: `$docUrl` should be quoted

scripts/e2e/start_api
```diff
@@ -5,19 +5,7 @@ grunt jshint
...
-		curl -X DELETE "$docUrl?rev=$rev"
-	fi
-}
-
-echo "[$0] deleting users created by test fixtures..."
-delete_doc "../_users/org.couchdb.user:bob"
-delete_doc "../_users/org.couchdb.user:clare"
-delete_doc "../_users/org.couchdb.user:chw-boss"
-delete_doc "../_users/org.couchdb.user:chw"
+scripts/delete_fictures
```
alxndrsn: Has this been tested? ;¬)

scripts/e2e/delete_fixtures
alxndrsn: Suggesting:

    local rev="$(curl "$docUrl" | jq --raw-output ._rev)"

Because it would be safer.  But I guess it's unlikely that doc URL would ever contain a space, so perhaps OK.

scripts/e2e/setup_fixtures
```diff
@@ -1,13 +1,44 @@
...
+delete_doc "fixture:user:chw-boss"
+
+echo "[$0] Finished cleanup. Creating new test data..."
+
+echo "[$0] creating the PARENT_PLACE..."
 curl -X PUT \
 	--data '{ "_id":"PARENT_PLACE", "type":"district_hospital", "name":"Big Parent Hospital" }' \
 	"$COUCH_URL/PARENT_PLACE"
 
-echo && echo "[$0] creating user: bob..."
```
alxndrsn: I guess you worked out what the extra echoes were for then :~)

scripts/e2e/setup_fixtures
```diff
@@ -1,13 +1,44 @@
...
+	local docUrl="$COUCH_URL/$1"
+	local rev="$(curl $docUrl | jq --raw-output ._rev)"
+	if [[ "$rev" != "null" ]]; then
+		curl -X DELETE "$docUrl?rev=$rev"
+	fi
+}
+
+delete_user() {
+	echo "[$0] Deleting user $1"
+	local username="$1"
```
alxndrsn: `username` is not used

scripts/e2e/delete_fixtures
alxndrsn: Also, it _is_ quoted 2 lines below

scripts/e2e/setup_fixtures
estellecomment: Whatever they were for, I can't see a difference either locally or in travis... I can research more if you think it's important, but that's enough for my curiosity.

scripts/e2e/delete_fixtures
estellecomment: oooh you can quote within quotes :woman_facepalming: 
I assumed you couldn't, and was trying to escape them, and so curl was complaining about `""http"`...

Thanks! Done.

PR #147, opened 2017-04-27
Allow collect to fetch XML forms

alxndrsn:
Fix for medic/medic-webapp#3390

### DIFF
--- a/controllers/forms.js
+++ b/controllers/forms.js
@@ -32,7 +32,7 @@ function listFormsXML(data, template, callback) {
   data.rows.forEach(function(row) {
     if (row.doc && row.doc._attachments && row.doc._attachments.xml) {
       ret.push(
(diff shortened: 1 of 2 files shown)


mandric:
also controllers/forms.js:80 should use http or https depending on the request.

```
template = 'http://%s/api/v1/forms/{{id}}.xml'.replace('%s', headers.host);
```

alxndrsn:
> I believe features and testing for a module should live together, so I expect to see tests and new functionality/fixes in the same repo/PR.

I will look at adding tests in the `medic-api` repo.  However, this code depends directly on the views defined in `medic-webapp`, and the dependency is defined on the webapp side, so the code under test is effectively in that repo rather than this.

> also controllers/forms.js:80 should use http or https depending on the request. … Maybe just open another issue for this.  

Will do.

PR #149, opened 2017-05-08
Webapp ddoc refactor

mandric:
# WIP - Phase 1 of kanso removal

- Removal of the kujua-sms kanso package dependency in medic-webapp
  - Add/migrate records creation support to medic-api directly
- Officially deprecate `/add` in medic-webapp return 50x and friendly error message.

This is part of a greater refactor to migrate functionality out of
CouchDB and kujua-sms kanso package.  This change migrates the
non-deprecated functionality used by the `/_design/medic/_rewrite/add`
path to medic-api directly.  There is no change in functionality here,
it should function as currently documented.

This refactor also includes new dependencies:

  https://github.com/medic/medic-smsparser
  A copy of medic-webapp/packages/kujua-sms/views/lib refactored as npm
  module.

  https://github.com/medic/medic-libphonenumber
  A copy of medic-webapp/packages/libphonenumber refactored as npm
  module.

  https://github.com/medic/medic-api-utils (might get renamed)
  A new place to put code that is shared by medic-api, medic-webapp and
  medic-sentinel.

I feel bad I didn't do any ES6 conversion here.

One potential issue, this module was duplicated/copied:

  - `medic-smsparser/tests/fixtures/form_definitions.js`
  - `medic-api/tests/unit/fixtures/form_definitions.js`

Issue: https://github.com/medic/medic-webapp/issues/3019


README.md
```diff
@@ -496,7 +498,7 @@ If required fields are not found return 500.
 
 If invalid JSON return error response 500.
 
-If submitting JSON and correspending form is not found on the server you will receive an error.
+If submitting JSON and corresponding form is not found on the server you will receive an error.
```
garethbowen: lol

controllers/record-utils.js
```diff
@@ -51,12 +70,240 @@ var createRecordByJSON = function(data, callback) {
...
+var getDataRecord = function(options, form_data) {
+
+  var form = options.form,
+      def = config.getForm(form);
+
+  var record = {
+   // _id: req.uuid,
+    type: 'data_record',
+    // todo maybe change libphonenumber to only take default_country_code
+    from: libphonenumber.normalize(config.settings, options.from) || options.from,
```
garethbowen: I'm still not comfortable with google-libphonenumber as a transitive dependency - I'd need to play with npm and shrinkwrap to ensure that bumping the version is as easy as before...

ddocs/Makefile
```diff
@@ -0,0 +1,18 @@
...
+	5984 compile medic-api/index.js > medic-api.json
+
+install: medic-api.json
+	# bug: does not fail on empty response
+	cat medic-api.json | \
+	  5984 fetch-revs "$(COUCH_URL)/medic" | \
+	  5984 bulk "$(COUCH_URL)/medic" | \
+	  tee /dev/stderr | \
+	  jq -e '.ok == true' > /dev/null
+
```
garethbowen: No, just, no. Don't use makefiles when we already use grunt. That's just confusing.

ddocs/Readme
```diff
@@ -0,0 +1,18 @@
...
+  http://docs.couchdb.org/en/2.0.0/install/unix.html
+  https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Releases/1.8.5
+ 
+To build the design doc run `make`.
+
+To install it on a running CouchDB run `make install`.
+
+Compiled design docs (JSON files) are committed to the repository and
+installed by medic-api during startup.
+
```
garethbowen: Put this readme in the root readme in its own section so it'll be easier to find.

ddocs/medic-api.json
```diff
@@ -0,0 +1 @@
... (userCtx && userCtx.roles) || [];\n  return roles.indexOf(role) >= 0;\n};\n\n/**\n * Update task/message object in-place.  Used by message update functions when\n * a message's state changes. Also adds new values to state history.\n *\n * @param {Object} task\n * @param {String} state\n * @param {Any} details (optional)\n * @api public\n */\nexports.setTaskState = function(task, state, details) {\n  task.state = state;\n  task.state_details = details;\n  task.state_history = task.state_history || [];\n  task.state_history.push({\n    state: state,\n    state_details: details,\n    timestamp: new Date().toISOString()\n  });\n};\n\n/*\n * Return task object that matches message uuid or a falsey value if match\n * fails.\n *\n * @api public\n */\nexports.getTask = function(uuid, doc) {\n  for (var i in doc.tasks) {\n    for (var j in doc.tasks[i].messages) {\n      if (uuid === doc.tasks[i].messages[j].uuid) {\n        return doc.tasks[i];\n      }\n    }\n  };\n  for (var i in doc.scheduled_tasks) {\n    for (var j in doc.scheduled_tasks[i].messages) {\n      if (uuid === doc.scheduled_tasks[i].messages[j].uuid) {\n        return doc.scheduled_tasks[i];\n      }\n    }\n  };\n};\n"}
```
garethbowen: Compiled code should be gitignored I think.

ddocs/medic-api/updates.js
```diff
@@ -0,0 +1,87 @@
...
+    msg = utils.getTask(data.message_id, doc);
+    if (!data.message_id || !msg) {
+        return fail('Message not found: ' + data.message_id);
+    }
+    utils.setTaskState(msg, data.state, data.details);
+    return [
+        doc,
+        success
+    ];
+};
```
garethbowen: I don't like having this as a couchdb update function. Why not expose it as an express API (if needed externally) or include it in the npm utils (if only needed internally)? Then we could get rid of the ddoc altogether I think?

server.js
```diff
@@ -146,7 +146,7 @@ app.get('/api/info', function(req, res) {
 app.get('/api/auth/:path', function(req, res) {
   auth.checkUrl(req, function(err, output) {
     if (err) {
-      return serverUtils.serverError(err, req, res);
+      return serverUtils.error(err, req, res);
```
garethbowen: Why did you change all these from serverError to error?

controllers/record-utils.js
mandric: One thing I wonder is if you keep the google-libphonenumber dep in medic-webapp as-is does anything change?  Either way we should have a way to patch the shrinkwrap config to make some exceptions for certain deps.  That's the real fix, right?

ddocs/Makefile
mandric: Yeah I felt the same way and started on the grunt path and after mucking around with config files for an hour or something and gave up.  This is my personal standard way to build ddocs not medic's and I agree it should get updated to Grunt but wanted to move on to more important things.  I do plan to fix this before merge though.

ddocs/medic-api.json
mandric: This particular asset should be provided as part of the distribution for easy installation IMO.  i.e. when medic-api starts it can just expect is there and use it during bootstrap. (todo)

ddocs/medic-api/updates.js
mandric: Because it's more efficient to update the doc this way, I thought it was useful.

server.js
mandric: Because the error messages were not being reported properly.  Afaict `serverUtils.error` is only for errors we are not handling.

controllers/record-utils.js
garethbowen: Currently to update the shrinkwrap all you have to do is `npm install google-libphonenumber@<x.y.z>`. With transitive dependencies I'm pretty sure that won't work so you'll have to update and publish medic-libphonenumber and then `npm install medic-libphonenumber@<x.y.z>` so a few extra steps to the current release process. It's possible that if `medic-libphonenumber` depends on a range then when we can update the transitive dependency with a standard `npm install` but I want to confirm this before we rely on having our own wrapper module. I really don't want to "patch the shrinkwrap" - that way leads to madness, because npm will blow away the patch any time shrinkwrap is updated.

ddocs/Makefile
garethbowen: Awesome, thanks!

ddocs/medic-api.json
garethbowen: There's a big difference between being included for distribution and being committed in to git. For example, our minified and browserified inbox.js script is distributed but it isn't committed into git. This resource is the same.

server.js
garethbowen: I think it would be better to fix the `serverError` implementation (or remove it altogether). It's quite confusing right now...

ddocs/medic-api/updates.js
mandric: Ok we're going for cleaner code and optimize to couchdb update later (minimizes http traffic) if we ever hit a performance issue, will refactor this bit.

server.js
mandric: I think it makes sense, one function reports errors details/messages and the other doesn't (I assume to avoid leaking sensitive data) but maybe the naming is confusing... I will have a look.

PR #152, opened 2017-05-12
3443 conflict hardening

SCdF:
NB: please don't squash


server.js
```diff
@@ -33,7 +33,7 @@ var _ = require('underscore'),
     monthlyRegistrations = require('./controllers/monthly-registrations'),
     monthlyDeliveries = require('./controllers/monthly-deliveries'),
     exportData = require('./controllers/export-data'),
-    messages = require('./controllers/messages'),
+    messages = require('./controllers/message-utils'),
```
alxndrsn: `message-utils` sounds like a weird name for a controller

server.js
```diff
@@ -199,11 +199,11 @@ app.get('/api/sms/', function(req, res) {
 app.get('/api/sms', function(req, res) {
   auth.check(req, 'can_access_gateway_api', null, function(err) {
     if (err) {
-      return serverUtils.error(err, req, res);
+      return serverUtils.error(err, req, res, true);
```
alxndrsn: Why do we want to `showPrompt` for these URLs?  AFAIK this is not required by gateway.

Or does this just mean "don't redirect to login page"?  If the latter, the parameter name in `serverUtils.error()` would benefit from being renamed.

tests/unit/controllers/sms-gateway.js
```diff
@@ -60,42 +62,44 @@ exports['post() should update statuses supplied in request'] = function(test) {
...
+  controller.post(req, err => {
     // then
-    test.equals(err, null);
-    test.equals(updateMessage.callCount, 5);
-    test.equals(updateMessage.withArgs('1', { state:'received-by-gateway' }).callCount, 1);
-    test.equals(updateMessage.withArgs('2', { state:'forwarded-by-gateway' }).callCount, 1);
-    test.equals(updateMessage.withArgs('3', { state:'sent' }).callCount, 1);
-    test.equals(updateMessage.withArgs('4', { state:'delivered' }).callCount, 1);
-    test.equals(updateMessage.withArgs('5', { state:'failed', details:{ reason:'bad' } }).callCount, 1);
+    test.equal(err, null);
```
alxndrsn: Are `equal` and `equals`...equivalent?

tests/unit/controllers/sms-gateway.js
```diff
@@ -123,125 +127,12 @@ exports['post() should provide WO messages in response'] = function(test) {
...
     });
-    test.equals(updateMessage.callCount, 3);
-    test.equals(updateMessage.withArgs('1', { state:'forwarded-to-gateway' }).callCount, 1);
-    test.equals(updateMessage.withArgs('2', { state:'forwarded-to-gateway' }).callCount, 1);
-    test.equals(updateMessage.withArgs('3', { state:'forwarded-to-gateway' }).callCount, 1);
-    test.done();
-  });
-};
-
-exports['post() should continue processing other stuff if saving a wt message fails' ] = function(test) {
```
alxndrsn: Are there tests somewhere for the new exception-throwing behaviour?

controllers/message-utils.js
```diff
@@ -1,20 +1,97 @@
...
+const _ = require('underscore'),
+      db = require('../db');
 
-var getTaskMessages = function(options, callback) {
+const getTaskMessages = function(options, callback) {
   db.medic.view('medic', 'tasks_messages', options, callback);
 };
 
+// copied from kujua-utils
+// TODO: get rid of this copy once Milan's refactor is complete
```
alxndrsn: If there's a ticket number for "milan's refactor" it would be great to include it with this comment

controllers/message-utils.js
```diff
@@ -1,20 +1,97 @@
...
+  task.state_history = task.state_history || [];
+  task.state_history.push({
+    state: state,
+    state_details: details,
+    timestamp: new Date().toISOString()
+  });
+};
+
+const getTaskForMessage = function(uuid, doc) {
+  const getTaskFromMessage = (tasks, uuid) => tasks.find(task => {
```
alxndrsn: i think either this should not be an inner-function, or you could remove the `uuid` param.

controllers/message-utils.js
```diff
@@ -1,20 +1,97 @@
...
+    }
+
+    const [task, docId] = getTaskAndDocForMessage(taskStateChange.messageId, docs);
+
+    if (!task) {
+      throw Error(`Message not found: ${taskStateChange.messageId}`);
+    }
+
+    fillTaskStateChangeByDocId(taskStateChange, docId);
+    setTaskState(task, taskStateChange.state, taskStateChange.details);
```
alxndrsn: It looks like these two functions (`fillTaskStateChangeByDocId()` and `setTaskState()`) should never be called independently of each other.  If that's the case, perhaps they should be combined, or at least scoped at the same level.

controllers/message-utils.js
```diff
@@ -1,20 +1,97 @@
...
+        return callback(err);
+      }
+      if (data.rows.length === 0) {
+        return callback({ code: 404, message: 'Not Found' });
+      }
+      callback(null, data.rows[0].value);
+    });
+  },
+  /*
+   * Gets all messages based on passed state
```
alxndrsn: I think the word `passed` is unnecessary here, and every time I read it I parse it as _past_, not _passed_.  Also, neither `state` nor `states` options are mandatory, so this comment may be misleading.

controllers/message-utils.js
```diff
@@ -51,45 +128,67 @@ module.exports = {
       callback(null, msgs);
     });
   },
-  updateMessage: function(id, body, callback) {
-    module.exports.getMessage(id, function(err, msg) {
+  /*
+   * See: updateMessageTaskStates
```
alxndrsn: Is this comment necessary?

controllers/message-utils.js
```diff
@@ -51,45 +128,67 @@ module.exports = {
...
-        callback(null, {
-          id: id,
-          success: data.payload.success
+
+        db.medic.bulk({docs: docs}, (err, results) => {
+          if (err) {
+            return callback(err);
+          }
+
+          const failures = results.filter((result) => !result.ok);
```
alxndrsn: Do we have a preferred style for including/excluding brackets for single-arg anonymous functions?  i.e.

    (result) => !result.ok

vs

    result => !result.ok

?

controllers/message-utils.js
```diff
@@ -51,45 +128,67 @@ module.exports = {
...
+          if (failures.length && !retriesLeft) {
+            const failure = `Failed to updateMessageTaskStates: ${JSON.stringify(failures)}`;
+            console.error(failure);
+            return callback(Error(failure));
+          }
+
+          if (failures.length) {
+            console.warn(`Problems with updateMessageTaskStates: ${JSON.stringify(failures)}`);
+            console.warn(`Retrying ${retriesLeft} more times`);
```
alxndrsn: Is there any point combining these log messages so they don't get split up by other messages in the log file?

controllers/message-utils.js
```diff
@@ -51,45 +128,67 @@ module.exports = {
...
+
+            const relevantChanges = _.chain(stateChangesByDocId)
+                                     .pick(_.pluck(failures, 'id'))
+                                     .values()
+                                     .flatten()
+                                     .value();
```
alxndrsn: very funcy

server.js
alxndrsn: IMO naming the file `controllers/messages.js` is better.

It seems reasonable to me that any controller could be accessed internally, as this should be roughly equivalent to accessing it via it's corresponding endpoint.  Of course, accessing a controller via `require()` will bypass auth, which may or may not be desirable in general.

controllers/sms-gateway.js
```diff
@@ -3,25 +3,27 @@
...
+      };
 
 function warn() {
-  var args = Array.prototype.slice.call(arguments, 0);
+  const args = Array.prototype.slice.call(arguments, 0);
   args.unshift('WARN', 'sms-gateway');
   console.error.apply(console, args);
 }
 
+// TODO: pull the logic out of CouchDB, make local, make bulk update
```
alxndrsn: Is this something to be done as part of this ticket?

controllers/sms-gateway.js
```diff
@@ -3,25 +3,27 @@
...
+        UNSENT: 'received-by-gateway',
+        PENDING: 'forwarded-by-gateway',
+        SENT: 'sent',
+        DELIVERED: 'delivered',
+        FAILED: 'failed',
+      };
 
 function warn() {
-  var args = Array.prototype.slice.call(arguments, 0);
+  const args = Array.prototype.slice.call(arguments, 0);
```
alxndrsn: Spread/unspread(?)

e.g.

```
function warn(...args) {
  console.error('WARN', ...args);
}
```

controllers/sms-gateway.js
```diff
@@ -3,25 +3,27 @@
...
   args.unshift('WARN', 'sms-gateway');
   console.error.apply(console, args);
 }
 
+// TODO: pull the logic out of CouchDB, make local, make bulk update
 function saveToDb(message, callback) {
   recordUtils.createByForm({
```
alxndrsn: Appreciate this isn't your code, but `createByForm()` seems to have very little relevance to what's happening here!

controllers/sms-gateway.js
```diff
@@ -30,73 +32,67 @@ function saveToDb(message, callback) {
...
-        // wasn't saved to DB, so message will stay in 'pending' status and be retried next time.
-        warn(err);
-      }
-      return callback();
-    }
-  );
+  const taskStateChanges = messages.map((message) => { return {
+    messageId: message.id,
+    state: 'forwarded-to-gateway'
+  };});
```
alxndrsn: 😖

alxndrsn: Would one of these be neater:
```
messages.map(message => ({ messageId: message.id, state: 'forwarded-to-gateway' }));
```
or
```
messages.map(message => ({
    messageId: message.id,
    state: 'forwarded-to-gateway'
}));
```
?

controllers/sms-gateway.js
```diff
@@ -30,73 +32,67 @@ function saveToDb(message, callback) {
...
-        to: message.to,
-        content: message.message,
-      };
-    });
-    markMessagesForwarded(messages, function() {
-      callback(null, messages);
-    });
+    (err, pendingMessages) => {
+      // TODO: I don't get this. Why wouldn't you return an error here? What
+      //       is the intention behind hiding it from the caller?
```
alxndrsn: I think we should be returning errors to the caller.  The original intention has been changed significantly, so there is no benefit to avoiding reporting this.

controllers/sms-gateway.js
```diff
@@ -30,73 +32,67 @@ function saveToDb(message, callback) {
...
+function addNewMessages(req, callback) {
   if (!req.body.messages) {
     return callback();
   }
-  async.eachSeries(req.body.messages, saveToDb, function(err) {
+
+  async.eachSeries(req.body.messages, saveToDb, err => {
     if (err) {
       warn(err);
     }
```
alxndrsn: As above, we probably want to `callback(err)` now.  Happy to consider it for a different ticket if you prefer.

controllers/message-utils.js
mandric: I think this is the main issue: https://github.com/medic/medic-webapp/issues/3019.  You might even note the new repo (currently `medic-api-utils` but that name is not final) where we plan to include shared code.

Gruntfile.js
```diff
@@ -96,6 +96,10 @@ module.exports = function(grunt) {
     'env:dev',
   ]);
 
+  grunt.registerTask('unit', [
+    'test_unit'
+  ]);
```
SCdF: It was to make it the same as webapp. We may want to, at some point, do a pass over our various grunt commands and make them the same everywhere.

controllers/message-utils.js
SCdF: Probably not. I was trying to be thorough because I found how `getMessages` wasn't the plural or `getMessage` really confusing.

controllers/message-utils.js
SCdF: I don't think we've settled on anything there yet. FWIW I like the latter one more, I'll change it.

(I actually wish you could make it even shorter with just one param and, like in other languages, e.g. `.filter(!_.ok)` or something.

controllers/message-utils.js
SCdF: Good idea. Done.

controllers/message-utils.js
SCdF: As long as you understand it :-)

controllers/sms-gateway.js
SCdF: Yep good idea. After other changes based on feedback on this PR, this function actually gets deleted :-)

controllers/sms-gateway.js
SCdF: It could be, but it's not required. I think Milan's refactor will make this easier. I'll also link this to the refactor ticket.

controllers/sms-gateway.js
SCdF: Yeah I don't get why it's by form either. This name should get refactored out once we've made it a bulk update, so I'm OK with it hanging around a little longer.

controllers/sms-gateway.js
SCdF: Yep! Didn't think to wrap it in `()`. Done.

controllers/sms-gateway.js
SCdF: Coo. I'll change it to `return callback(err)`.

controllers/sms-gateway.js
SCdF: That's cool, it's a small change. Done.

server.js
SCdF: Yeah I planned on renaming it, I just never got around to it I guess. I think I didn't want it in the same commit. I'll rename it now.

server.js
SCdF: It's definitely weird. I got rid of it for these calls. The documentation implies I should have it, but looking at the code it's not necessary (IMO in any situation).

tests/unit/controllers/sms-gateway.js
SCdF: Yes. Equals (and any other puraly sounding functions) is actually deprecated and has been for some time. I have considered changing them everywhere in the code but I don't want to piss people off ;-)

tests/unit/controllers/sms-gateway.js
SCdF: Nope. Added.

controllers/message-utils.js
alxndrsn: I never said that!  TBH I had to think about it a bit, and I guess the input is a bit weird to require the `flatten()`, but I think it's good.

PR #156, opened 2017-05-25
Split couch _changes requests with _doc_ids filter into batches of 100

alxndrsn:
still working on fixing tests

### DIFF
--- a/handlers/changes.js
+++ b/handlers/changes.js
@@ -1,4 +1,5 @@
 var _ = require('underscore'),
+    async = require('async'),
     auth = require('../auth'),
     config = require('../config'),
     serverUtils = require('../server-utils'),
@@ -8,6 +9,13 @@ var _ = require('underscore'),
     CONTACT_TYPES = ['person', 'clinic', 'health_center', 'district_hospital'],
     inited = false,
     continuousFeeds = [];
+
+/**
+ * As per https://issues.apache.org/jira/browse/COUCHDB-1288, there is a 100 doc
+ * limit on processing changes feeds with the _doc_ids filter within a
+ * reasonable amount of time.
+ */
+const MAX_DOC_IDS = 100;
 
 var error = function(code, message) {
   return JSON.stringify({ code: code, message: message });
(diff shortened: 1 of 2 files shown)


handlers/changes.js
```diff
@@ -161,15 +169,21 @@ var cleanUp = function(feed) {
...
+  var allIds = _.union(feed.requestedIds, feed.validatedIds);
+  var chunks = [];
+  while (allIds.length) {
+    chunks.push(allIds.splice(0, MAX_DOC_IDS));
+  }
+
   // we cannot call 'changes' in nano because it only uses GET requests and
   // our query string might be too long for GET
-  feed.changesReq = db.request({
+  async.parallel(chunks.map(docIds => (callback) => db.request({
```
SCdF: Thinking we might want to not do parallel here, so that we stress the server less? I guess it really depends on what that 100 limit is guarding against.

SCdF: Also, currying, very fancy!

handlers/changes.js
```diff
@@ -178,19 +192,48 @@ var getChanges = function(feed) {
...
       // Attempting to log out the response usefully to see what's occuring
-      var malformedChangesError = 'No _changes error, but malformed response: ';
-      var printableChanges = JSON.stringify(changes);
-      console.error(malformedChangesError, printableChanges);
-      feed.res.write(error(503, malformedChangesError + printableChanges));
+      console.error('No _changes error, but malformed response:', JSON.stringify(changes));
+      err = true;
     } else {
-      prepareResponse(feed, changes);
+      changes.results.forEach(r => merged.results.push(r));
```
SCdF: Can this not just be a concat?

handlers/changes.js
```diff
@@ -178,19 +192,48 @@ var getChanges = function(feed) {
...
+      merged.last_seq = Math.max(merged.last_seq, changes.last_seq);
     }
-    feed.res.end();
   });
+
+  if (err) {
+    return;
+  }
+
+  merged.results.sort((a, b) => a.seq - b.seq);
```
SCdF: Probably want another todo here about couchdb 2.0. It will mostly be the same, just need to cast to an int and get rid of the string after the number

handlers/changes.js
alxndrsn: Yeah i was wondering about that - so we would just do
```
merged.results = merged.results.concat(changes.results)
```
Certainly sounds like it should be quicker.

handlers/changes.js
alxndrsn: i'm surprised this didn't require more brackets

handlers/changes.js
SCdF: Yep I think that should do the trick.

handlers/changes.js
alxndrsn: In fact it may have too many already

SCdF: I'm actually not sure you need the ones around `callback`…

alxndrsn: `async.series`?  TBH I don't think we should be worried about server load.

SCdF: That would be the other option, yeah. I'm not sure, I guess we can try both and see how it runs.

handlers/changes.js
alxndrsn: Is chopping the string part definitely valid?

PR #160, opened 2017-05-31
Updates people and places APIs to minify contacts

garethbowen:
The people and places APIs were missed from the Linked Contacts refactor so were still creating inlined parents and contacts.

medic/medic-webapp#2635

### DIFF
--- a/controllers/people.js
+++ b/controllers/people.js
@@ -1,16 +1,16 @@
-var _ = require('underscore'),
-    db = require('../db'),
-    utils = require('./utils'),
-    places = require('./places');
+const _ = require('underscore'),
+      db = require('../db'),
+      utils = require('./utils'),
+      places = require('./places');
 
-var getPerson = function(id, callback) {
-  var error = function(msg, code) {
+const getPerson = (id, callback) => {
+  const error = (msg, code) => {
     return callback({
       code: code || 404,
       message: msg || 'Failed to find person.'
     });
   };
-  db.medic.get(id, function(err, doc) {
+  db.medic.get(id, (err, doc) => {
     if (err) {
       if (err.statusCode === 404) {
         return error();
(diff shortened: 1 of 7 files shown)


garethbowen:
@mandric Review if you please? There's a heap of refactoring so you're better off reviewing each commit separately.

mandric:
It's worth noting here how we use the word "refactor".  Refactor to me means major code logic changes or redesigning code and has nothing to do with reformatting or style (ES6 shorthand).  Also when you use "refactor" in the title of a commit I'd expect a body to explain in plain english what the refactor is, highlighting the rationale behind the refactor.  So I was immediately confused by two things in the first commit, straight away, simply because of a terminology issue.  It might be worth documenting our definitions for "reformat" and "refactor".  I hate to get bogged down in minor process adjustments but you put in effort to split up the commits nicely, maybe we should get the wording down and crystalize it!

controllers/places.js
```diff
@@ -252,3 +269,4 @@ module.exports.createPlace = createPlaces;
 module.exports.getPlace = getPlace;
 module.exports.updatePlace = updatePlace;
 module.exports.getOrCreatePlace = getOrCreatePlace;
+module.exports.minify = minify;
```
mandric: Are we still doing the export using `_` (underscore prefix) if it's a private function thing (only exported for testing).

controllers/people.js
```diff
@@ -24,12 +24,10 @@ var getPerson = function(id, callback) {
   });
 };
 
-var isAPerson  = function(obj) {
-  return obj.type === 'person';
-};
+const isAPerson = obj => obj.type === 'person';
```
garethbowen: One liners auto-return.

controllers/places.js
garethbowen: `minify` is used outside this module so it shouldn't have the underscore. Soon we can extract it into your library but for now this is a decent place to export it from.

IMO we should stop using the underscore prefix convention. My preferred way is to break each module up so it's small enough to unit test without having to expose the guts of it. If that's not practical then the next best way is to export more function when the TEST_ENV is set. That way you can't accidentally use one in production code.

garethbowen:
To me, refactoring is simply any change that doesn't change the external behaviour.

PR #161, opened 2017-06-02
Require that CouchDB requires auth

SCdF:
To prevent accidents, make sure that when api boots CouchDB is not
running the hilariously insecure default "admin party" mode, where no
authentication is required.

### DIFF
--- a/server.js
+++ b/server.js
@@ -712,14 +712,19 @@ const envVarsCheck = callback => {
 };
 
 const couchDbNoAdminPartyModeCheck = callback => {
-  const noAuthUrl = process.env.COUCH_URL.replace(/\/\/[^@]+@/, '//'),
-        protocol = noAuthUrl.split('://')[0];
-  const net = require(protocol);
-  net.get(noAuthUrl, err => {
-    if (err && err.statusCode === 401) {
+  const url = require('url'),
+        noAuthUrl = url.parse(process.env.COUCH_URL),
+        protocol = noAuthUrl.protocol.replace(':', ''),
+        net = require(protocol);
+
+  delete noAuthUrl.auth;
+
+  net.get(url.format(noAuthUrl), ({statusCode}) => {
+    // We expect to be rejected because we didn't provide auth
+    if (statusCode === 401) {
       callback();
     } else {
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -711,6 +711,19 @@ const envVarsCheck = callback => {
   }
 };
 
+const couchDbNoAdminPartyModeCheck = callback => {
+  const noAuthUrl = process.env.COUCH_URL.replace(/\/\/[^@]+@/, '//'),
+        protocol = noAuthUrl.split('://')[0];
```
garethbowen: It'd be great if we could start to move away from regexing URLs as it's hard to read and potentially error prone. How about using the URL api? Something like this...

```js
const url = require('url');
const noAuthCouchUrl = new url(process.env.COUCH_URL);
noAuthCouchUrl.username = null;
noAuthCouchUrl.password = null;
const net = require(noAuthCouchUrl.protocol);
net.get(noAuthUrl.toString(), err => {

```

server.js
```diff
@@ -711,6 +711,19 @@ const envVarsCheck = callback => {
...
+  const net = require(protocol);
+  net.get(noAuthUrl, err => {
+    if (err && err.statusCode === 401) {
+      callback();
+    } else {
+      callback(err || new Error('CouchDB is configured in admin-party mode (ie no auth required)'));
```
garethbowen: Maybe create a medic-docs document for how to get out of admin-party mode and link from here? Or link to 3rd party docs if you can find something good enough.

garethbowen:
I love the pre-flight checklist concept, and I'll add a couple of things when I find time. Ideally this would also be exposed in the UI somewhere so users would get a useful message rather than a 502, but one step at a time!

server.js
SCdF: I was originally like this (well, the real code is grosser than this because the pre 8.x url api is not great), but then I got rid of it for some reason. I can't remember why though. I'll change it back and see if whatever issue I had before comes up.

server.js
SCdF: There is tonnes of CouchDB docs on it, I'll link one.

PR #163, opened 2017-06-13
Adds support for storing read status in meta db

garethbowen:
Adds a migration to extract the read status from the docs and
into the meta db.

Adds an endpoint to handle the lazy creation of the user meta
db.

Also improves the migration output to be formatted rather than
stringifying the Error.

medic/medic-webapp#2418

### DIFF
--- a/controllers/create-user-db.js
+++ b/controllers/create-user-db.js
@@ -12,7 +12,7 @@ const createDatabase = (username, callback) => {
     const params = {
       db: dbName,
       path: '/_security',
-      method: 'put',
+      method: 'PUT',
       body: {
         admins: { names: [ username ], roles: [] },
         members: { names: [], roles:[] }
--- a/controllers/create-user-db.js
+++ b/controllers/create-user-db.js
@@ -1,26 +1,5 @@
-const db = require('../db'),
-      auth = require('../auth');
-
-const getDbName = username => `medic-user-${username}-meta`;
-
-const createDatabase = (username, callback) => {
-  const dbName = getDbName(username);
-  db.db.create(dbName, err => {
-    if (err) {
-      return callback(err);
-    }
-    const params = {
-      db: dbName,
-      path: '/_security',
-      method: 'PUT',
-      body: {
-        admins: { names: [ username ], roles: [] },
-        members: { names: [], roles:[] }
-      }
-    };
-    db.request(params, callback);
-  });
-};
+const auth = require('../auth'),
+      userDb = require('../lib/user-db');
 
 const checkPermissions = (req, callback) => {
   auth.getUserCtx(req, (err, userCtx) => {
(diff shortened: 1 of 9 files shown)


migrations/extract-read-status.js
```diff
@@ -0,0 +1,87 @@
...
+    db.request(params, callback);
+  });
+};
+
+const ensureDbExists = (username, dbName, callback) => {
+  db.db.get(dbName, err => {
+    if (err && err.statusCode === 404) {
+      return createDb(username, dbName, callback);
+    }
+    callback();
```
SCdF: Do you want to be checking for just `err` when it's not a 404 as well?

db.js
```diff
@@ -116,6 +117,31 @@ if (couchUrl) {
...
+      }
+      console.log(`        Processing ${skip} to ${skip + batchSize} docs of ${response.total_rows} total`);
+      iteratee(response.rows, err => {
+        const keepGoing = response.total_rows > (skip + batchSize);
+        callback(err, keepGoing);
+      });
+    });
+  };
+
+  module.exports.batch = (query, iteratee, batchSize, callback) => {
```
SCdF: It's also worth considering that if we're having a technique elevated into the "standard library" (as it were), we probably want to do skipping the more correct / performant way, using keys, as opposed to numeric record skipping.

controllers/create-user-db.js
```diff
@@ -0,0 +1,46 @@
...
+const createDatabase = (username, callback) => {
+  const dbName = getDbName(username);
+  db.db.create(dbName, err => {
+    if (err) {
+      return callback(err);
+    }
+    const params = {
+      db: dbName,
+      path: '/_security',
+      method: 'put',
```
garethbowen: Done.

migrations/extract-read-status.js
garethbowen: Yeah. I'm not sure what that `err` would be at this stage if not a 404, but I've modified the code to pass it to the callback so at least we'll see it reported.

db.js
garethbowen: Yes probably but it's complicated so would need extensive testing. I've raised a separate issue to investigate this: medic/medic-webapp#3553

lib/user-db.js
```diff
@@ -0,0 +1,56 @@
...
+
+const putDdoc = (dbName, callback) => {
+  db.use(dbName).insert(ddoc, callback);
+};
+
+module.exports = {
+  getDbName: username => `medic-user-${username}-meta`,
+  create: (username, callback) => {
+    const dbName = module.exports.getDbName(username);
+    createDb(dbName, err => {
```
SCdF: Can I getta `async.series`. Gotta kill the JS callback hate pyramid

db.js
```diff
@@ -116,6 +117,42 @@ if (couchUrl) {
...
+        callback(err, keepGoing);
+      });
+    });
+  };
+
+  /**
+   * Run an operation over all documents returned from the query in batches.
+   *
+   * query function     Called with a skip number and a callback to
+   *                    invoke with the next batch.
```
garethbowen: Done.

lib/user-db.js
garethbowen: Here's your `async.series` and I threw in a few `async.apply` calls for free!

PR #165, opened 2017-06-19
Handle upgrades in API instead of dashboard

alxndrsn:


alxndrsn:
@estellecomment @SCdF any thoughts on what we should do with this?

SCdF:
@alxndrsn I think we should merge it into master. If we're not comfortable having this around when we don't officially support it, we should merge it but hide it behind a feature toggle.

SCdF:
@alxndrsn I've created this PR which is just the relevant changes: https://github.com/medic/medic-api/pull/187 I'll work on cleaning it up alongside https://github.com/medic/medic-webapp/pull/3750

If you think this is a good base we can delete this branch / pr

SCdF:
Superceeded by https://github.com/medic/medic-api/pull/187

PR #168, opened 2017-06-26
Improves pagination strategy of bulk get

garethbowen:
Pulls out the db.batch function into lib/db-batch

Uses the startkey strategy instead of skip for improved
performance and removing the chance of missing records due to
view reordering.

medic/medic-webapp#3553

### DIFF
--- a/lib/db-batch.js
+++ b/lib/db-batch.js
@@ -1,0 +1,55 @@
+const async = require('async'),
+      _ = require('underscore'),
+      db = require('../db'),
+      DEFAULT_BATCH_LIMIT = 100; // 100 is a good compromise of performance and stability
+
+const runBatch = (ddocName, viewName, viewParams, iteratee, callback) => {
+  db.medic.view(ddocName, viewName, viewParams, (err, response) => {
+    if (err) {
+      return callback(err);
+    }
+    console.log(`        Processing doc ${response.offset}`);
+    let nextPage;
+    if (response.rows.length === viewParams.limit) {
+      const lastRow = response.rows.pop();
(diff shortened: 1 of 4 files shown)


lib/db-batch.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+/**
+ * Run an operation over all documents returned from the query in batches.
+ * TODO fix this - now completely wrong
+ *
+ * ddocName (string)    Name of the ddoc the view is defined in.
+ * viewName (string)    Name of the view.
+ * viewParams (object)  Parameters to pass to the view query.
+ *  - `limit` defaults to 100 and can be overriden.
```
SCdF: My eyes flick downward
Line forty two contradicts
Do I say something?

lib/db-batch.js
```diff
@@ -0,0 +1,55 @@
...
+ *  - `include_docs` defaults to `true` and cannot be overriden.
+ *  - `startkey` and `startkey_docid` cannot be overriden.
```
SCdF: True after one batch
Not true at the start though
Not sure it matters

lib/db-batch.js
```diff
@@ -0,0 +1,55 @@
...
+    const docs = response.rows.map(row => row.doc);
+    iteratee(docs, err => {
+      callback(err, nextPage);
+    });
+  });
+};
+
+/**
+ * Run an operation over all documents returned from the query in batches.
+ * TODO fix this - now completely wrong
```
garethbowen: Documentation
So easy to mess it up
I'll commit a fix

PR #170, opened 2017-06-29
Return the correct error message medic/medic-webapp#3587

SCdF:
If you try to destructure undefined it gives you a really confusing
error message:
TypeError: Cannot read property 'Symbol(Symbol.iterator)' of undefined

Instead, if we can't find the task return null, and have our correct
existing error message handle it

### DIFF
--- a/controllers/messages.js
+++ b/controllers/messages.js
@@ -59,13 +59,13 @@ const applyTaskStateChangesToDocs = (taskStateChanges, docs) => {
 
   taskStateChanges.forEach(taskStateChange => {
     if (!taskStateChange.messageId) {
-      throw Error('Message id required');
+      throw {code: 400, message: 'Message id required'};
     }
 
     const {task, docId} = getTaskAndDocForMessage(taskStateChange.messageId, docs);
 
     if (!task) {
-      throw Error(`Message not found: ${taskStateChange.messageId}`);
+      throw {code: 404, message: `Message not found: ${taskStateChange.messageId}`};
     }
 
     fillTaskStateChangeByDocId(taskStateChange, docId);
--- a/controllers/messages.js
+++ b/controllers/messages.js
@@ -39,6 +39,8 @@ const getTaskAndDocForMessage = function (messageId, docs) {
       return [task, doc._id];
     }
   }
+
+  return [null,null];
 };
 
 /*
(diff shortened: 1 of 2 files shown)


controllers/messages.js
```diff
@@ -39,6 +39,8 @@ const getTaskAndDocForMessage = function (messageId, docs) {
       return [task, doc._id];
     }
   }
+
+  return [null,null];
```
garethbowen: This is really ugly... why are we returning an array when the two items aren't the same? How about returning `{ task: task, _id: doc._id }`, and `{ task: null, _id: null }`? You can still destructure it on the other side with `const { task, docId }` but each component is labelled and obvious.

SCdF:
@garethbowen on HTTP 400: "The HTTP 400 Bad Request response status code indicates that the server could not understand the request due to invalid syntax". The syntax is not invalid here, we just can't find the document the message is attached to.

This also may be heretical, but does labouring over HTTP codes to find the right one matter that much? What different decision is the client going to make based on the HTTP code?

To avoid being burnt as an unbeliever, here are my best guesses to the correct HTTP codes for the three specific errors that code can generate:
 - "Message id required": closest I can see is a 400
 - "Message not found": sounds like a 404, but that makes me feel uncomfortable because that has its own mythology around it
 - "Failed to updateMessageTaskStates": should be the same as all the passively dealt with PouchDB errors, which end up bubbling up to be 500s (so no change).

controllers/messages.js
SCdF: I'll change it

SCdF: (Interesting aside: if you don't return `{}` in this instance you actually get a slightly nicer error message too: `Cannot match against 'undefined' or 'null'`. That error is at least understandable)

garethbowen:
@SCdF I totally agree that we shouldn't worry too much about status codes because we don't know of anyone relying on them. The important thing is making is easy to diagnose what's going on. If I curled and got a 50x I'd assume we had a bug in API and start down that rabbit hole. Also note if you follow the code down the response is prefixed with "Server error: ". And finally, when a server error is detected the stack is logged to console.

This is all appropriate when an unexpected error has occurred but highly misleading when the user input is incorrect. Your proposed status codes are fine.

I think I'll raise this in the meeting because it's not the first time this has come up.

PR #171, opened 2017-07-03
Check content type

SCdF:
If you omit the content type body gets marshalled as {}, which is
very confusing and hard to spot, because this endpoint does multiple
things, and so will cheerfully ignore your body and continue on.


SCdF:
Wow I completely forgot about this and thought it got merged already. Your PR looks better than my PR. I'll close this one and we'll go with yours.

PR #180, opened 2017-07-14
Better error reporting medic/medic-webapp#3656

SCdF:
# Description

Check from and content ourselves so we complain about the correct
names before they get remapped to something else.

Dump out the request body if there is an error.

medic/medic-webapp#3656

# Review checklist


SCdF:
Actually we don't want to do this, we want to let empty SMS through. I'll raise another PR.

PR #183, opened 2017-07-18
Reject non-JSON requests for json-only endpoints

alxndrsn:
# Description

ref https://github.com/medic/medic-api/pull/171

# Review checklist

- [x] Readable: Concise, well named, follows the [style guide](https://github.com/medic/medic-docs/blob/master/development/style-guide.md), documented if necessary.
- [x] Documented: Announced in Changes.md in plain English. Configuration and user documentation on [medic-docs](https://github.com/medic/medic-docs/)
- [x] Internationalised: All user facing text
- [x] Backwards compatible: Works with existing data and configuration or includes a migration. Any breaking changes documented in Changes.md.


alxndrsn:
Test added at https://github.com/medic/medic-webapp/pull/3682

alxndrsn:
@SCdF it wouldn't be hard to extend this test to check a longer list of endpoints, if that is deemed useful.

Gruntfile.js
```diff
@@ -56,7 +57,10 @@ module.exports = function(grunt) {
         }
       },
       e2e: {
-        src: ['tests/e2e/**/*.js'],
+        src: [
+          'tests/e2e/*.spec.js',
+          'tests/e2e/**/*.js',
```
SCdF: Why is this changing? What does `*.spec.js` mean over `**/*.js`? Doesn't the second one encapsulate the first one?

tests/e2e/server.spec.js
```diff
@@ -0,0 +1,33 @@
+const assert = require('chai').assert;
+const request = require('request-promise-native');
+
+describe('server', () => {
+
+  describe('JSON-only endpoints', () => {
+
+    it('should require application/json Content-Type header', () => {
```
SCdF: Nice one!

Can we have one more for the positive case please?

alxndrsn: I'd hope the positive path is already tested in function-specific tests...

Gruntfile.js
alxndrsn: In https://github.com/medic/medic-api/tree/master/tests/e2e, there are utility classes which will match `*.js`.

`.spec.` is meant to signify that the file in question is a test file.  `.spec.js` is a fairly common convention for test file names.  [In the mocha docs](https://mochajs.org/#browser-specific-methods) they use `test.whatever.js`.  I don't mind if we drop it, and files without `describe()` calls are probably ignored by the test runner, but it seems useful to me to have some prefix or suffix convention to denote files which contain tests rather than just have test-related code in them.

alxndrsn:
@SCdF Can this be merged?

alxndrsn:
@garethbowen I'd appreciate your thoughts on this in Stefan's absence.

Gruntfile.js
garethbowen: I like the `.spec.js` convention and Bede has started using it in webapp so stick with it. I agree with Stefan - I thought `tests/e2e/**/*.js` would also pick up your new test but I assume you tried and it didn't work, so no problem adding it.

server.js
```diff
@@ -52,6 +52,25 @@ var _ = require('underscore'),
...
+        code: 400,
+        message: 'Content-Type must be application/json'
+      }, req, res);
+    } else {
+      callback(req, res);
+    }
+  });
+};
+['delete', 'post', 'put'].forEach(method => {
+  app[method + 'Json'] = (path, callback) => {
```
garethbowen: This makes it _really_ hard to follow when debugging because I won't be able to search for the "postJson" string. Can you change this to...

```js
['deleteJson', 'postJson', 'putJson'].forEach(method => {
  app[method] = (path, callback) => {
```

It's only costs a few characters but is much easier to follow IMO.

alxndrsn:
Merged manually.

PR #184, opened 2017-07-26
Silently fail to save sms without content or from fields

henokgetachew:
# Description

Fixes this error: https://github.com/medic/medic-projects/issues/2480#issuecomment-317370396.
This PR is an extension to a change @SCdF made in this issue: medic/medic-webapp/issues/3672.  The error is logged as per the change by @SCdF, this PR just makes the saving silently fail if there are no `from` and `content` fields - this error is logged already by the changes that Stefan made in the other PR.

medic/medic-webapp#[3672]

# Review checklist

### DIFF
--- a/controllers/sms-gateway.js
+++ b/controllers/sms-gateway.js
@@ -23,13 +23,15 @@ const _ = require('underscore'),
 // See: https://github.com/medic/medic-webapp/issues/3019
 // Specifically, this should be in a new repo that we can pull in via npm
 function saveToDb(message, callback) {
-  if(message.from && message.content){
-    recordUtils.createByForm({
-      from: message.from,
-      message: message.content,
-      gateway_ref: message.id,
-    }, callback);
+  if(!message.from || !message.content){
+    return callback();
   }
+
+  recordUtils.createByForm({
+    from: message.from,
+    message: message.content,
+    gateway_ref: message.id,
+  }, callback);
 }
 
 function mapStateFields(update) {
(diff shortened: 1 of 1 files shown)


mandric:
The client can choose to ignore responses but the server should still provide useful responses and logging.  That patch/commit on medic-api looks like a bug to me.  Not sure how any of this helps the current problems (issue at hand) but that's just my $.02.  I would try to add the exception throw/error returning back to medic api and fix gateway.  This looks like two hacks now instead of a fix.

henokgetachew:
That would return it to the original design @mandric. Basically removing both @SCdF and my changes would take you back there. That would mean we're fixing this on the side of gateway - my question is: is that practical? Given that the gateway is located temotely and that both upgrades and retrieving the logs as well as monitoring the logs is not easy (?)

henokgetachew:
I just realized that I was trying to do the opposite of what Stefan started doing, which is to save the empty message - titled [here](https://github.com/medic/medic-api/pull/181/files/32ca364728ff6c3ea985ba1571f70f9017422cab): as Allow for empty sms.  Therefore I'm just passing an empty string as the content of the message if the message doesn't have value.  @mandric let me know if that will do until we come up with a better error reporting mechanism for this issue.

alxndrsn:
@henokgetachew I'm surprised you're seeing cases where gateway is failing to provide either `content` or `from`.  I'd be really interested on looking closer at this with you.

alxndrsn:
https://github.com/medic/medic-api/pull/185 may supersede this PR.

henokgetachew:
Closing this as `master` didn't have the most recent commits and there was a commit that was in 2.12.1 that wasn't in `master`.

PR #185, opened 2017-07-27
SMSs with empty from and content fields are still valid messages.  They should not be rejected by API.  If they are unexpected or unwanted, this should beBe more precise and permissive with SMS field requirements

alxndrsn:
SMSs with empty from and content fields are still valid messages.  They should not be rejected by API.  If they are unexpected or unwanted, API should discard them and log this at its discretion.

Also, the error thrown in `sms-gateway.saveToDb()` comes from a commit which appears on `2.12.x` branch, but not on `master`.  Is that deliberate, or should it (and this PR) be ported to `master`?

### DIFF
--- a/controllers/record-utils.js
+++ b/controllers/record-utils.js
@@ -22,7 +22,8 @@ var request = function(opts, callback) {
 };
 
 var createByForm = function(data, callback) {
-  if (empty(data.from)) {
+  // We're OK with from being empty - sometimes weird things happen with SMS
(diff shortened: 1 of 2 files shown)


mandric:
I don't really agree with this change but If we are changing `from` to be optional (empty string), then we should update the docs. 
 https://github.com/medic/medic-api#form-parameters

SCdF:
I have also cherry picked this into 2.13.x and master

PR #187, opened 2017-08-02
Horticulturalist upgrade support

SCdF:
# Description

API now knows how to stage a ddoc from our builds repo for Horticulturalist to then pick up and deploy.

medic/medic-webapp#3749

# Review checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1237,3 +1237,32 @@ DELETE /api/v1/users/mary
 ```
 HTTP/1.1 200 OK
 ```
+
+# Upgrades
+
+Initiates an upgrade to the provided version. Once this endpoint is called [Horticulturalist](https://github.com/medic/horticulturalist) will take over and complete the installation.
+
+### Permissions
+
+`_admin` (CouchDB Administrator)
+
+## Example
+
+```
+POST /api/v1/upgrade
+{
+  "build": {
+    "namespace": "medic",
+    "application": "medic",
+    "version": "3.0.0-beta.1"
+  }
+}
+```
+
+For potential forwards compatibility, you must pass the `namespace` and `application` as `medic`.
+
+The `version` should correspond to a release, pre-release or branch that has been pushed to our builds server (currently hard-coded to https://staging.dev.medicmobile.org/builds). This happens automatically upon a successful travis run.
+
+Calling this endpoint will potentially restart API.
+
+It is expected that the caller ensures forwards or backwards compatibility is maintained between deployed versions. This endpoing does not stop you from "upgrading" to an earlier version, or a branch that is incompatible with your current state.
--- a/controllers/upgrade.js
+++ b/controllers/upgrade.js
@@ -35,13 +35,6 @@ module.exports = (buildInfo, username) => {
     });
   }
 
-  if (true) {
-    console.log('YYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYY');
-    console.log(buildInfo, username);
-    console.log('YYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYY');
-    return Promise.resolve();
-  }
-
   console.log('upgrade()', `Upgrading to ${JSON.stringify(buildInfo)}…`);
 
   console.log('upgrade()', 'Fetching newDdoc…');
--- a/controllers/upgrade.js
+++ b/controllers/upgrade.js
@@ -70,6 +70,11 @@ module.exports = (buildInfo, username) => {
           };
 
           console.log('upgrade()', `Staging new ddoc as ${newDdoc._id}`);
+          // TODO: if we have already pushed a staged ddoc this can cause a 409
+          //       Do we want to deal with this? Do we want to delete the existing
+          //       ddoc first? What if horti is in the middle of dealing with it?
+          //       Horti deletes this document once it's copied it into the right
+          //       place, so maybe leave this / make the error messag clearer
           return targetDb.put(newDdoc);
         })
         .then(() => console.log('upgrade()', 'newDdoc uploaded, awaiting Horticulturalist'));
--- a/controllers/upgrade.js
+++ b/controllers/upgrade.js
@@ -6,40 +6,55 @@ const buildDb = new PouchDB(buildDbUrl);
 const targetDbUrl = process.env.COUCH_URL;
 const targetDb = new PouchDB(targetDbUrl);
 
-module.exports = (version, username) => {
-  console.log('upgrade()', `Upgrading to ${version}…`);
+// TODO: formalise this (TODO add ticket)
+const stagedDdocName = name => name + ':staged';
+
+// TODO: formalise all of this data munging (TODO add ticket)
+const ddocName = buildInfo =>
+ `${buildInfo.namespace}:${buildInfo.application}:${buildInfo.version}`;
+
+module.exports = (buildInfo, username) => {
+  // While we've set the api to in theory be more generic, this is for if we
+  // split this code out (and say, put it with horticulturalist). For now we
+  // just support medic deploying medic :-)
+  if (buildInfo.namespace !== 'medic' ||
+      buildInfo.application !== 'medic') {
+    throw new Error(
+      `We only support medic-webapp right now, not this: ${buildInfo}'`);
+  }
+
+  console.log('upgrade()', `Upgrading to ${JSON.stringify(buildInfo)}…`);
 
   console.log('upgrade()', 'Fetching newDdoc…');
   return buildDb
-    .get(version, { attachments:true })
-    .then(newDdoc => console.log('upgrade()', 'Fetched newDdoc.') || newDdoc)
+    .get(ddocName(buildInfo), { attachments:true })
+    .then(newDdoc => {
+      console.log('upgrade()', 'Fetched newDdoc.');
 
-    .then(newDdoc => console.log('upgrade()', 'Fetching oldDdoc…') || newDdoc)
-    .then(newDdoc =>
-      targetDb.get('_design/medic')
-        .then(oldDdoc => console.log('upgrade()', 'Fetched oldDdoc.') || oldDdoc)
-
+      console.log('upgrade()', 'Fetching oldDdoc…');
+      return targetDb.get('_design/medic')
         .then(oldDdoc => {
-          newDdoc.app_settings = oldDdoc.app_settings;
-          newDdoc._id = oldDdoc._id;
+          console.log('upgrade()', 'Fetched oldDdoc.');
+
+          newDdoc._id = stagedDdocName(oldDdoc._id);
           newDdoc._rev = oldDdoc._rev;
 
           newDdoc.deploy_info = {
             timestamp: new Date().toString(),
             user: username,
-            version: version,
+            version: buildInfo.version,
           };
 
-          console.log('upgrade()', 'Uploading new ddoc…');
-          return targetDb.put(newDdoc)
-            .then(ret => console.log('upgrade()', 'newDdoc uploaded.') || ret);
-
-        }))
-    .catch(err => {
-      if (err.status === 404) {
-        err = new Error(`Version not found: ${version}`);
-        err.expected = true;
-      }
-      throw err;
-    });
+          console.log('upgrade()', 'Uploading new ddoc into staging position');
+          return targetDb.put(newDdoc);
+        })
+        .then(() => console.log('upgrade()', 'newDdoc uploaded, awaiting Horticulturalist'));
+      })
+      .catch(err => {
+        if (err.status === 404) {
+          err = new Error(`Version not found: ${buildInfo.version}`);
+          err.expected = true;
+        }
+        throw err;
+      });
 };
--- a/controllers/upgrade.js
+++ b/controllers/upgrade.js
@@ -4,7 +4,7 @@ PouchDB.plugin(require('pouchdb-adapter-http'));
 // const buildDbUrl = 'https://staging.dev.medicmobile.org/_couch/builds';
 
 //////////////////////////////// HACK
-const buildDbUrl = 'https://admin:pass@localhost:5984/builds';
+const buildDbUrl = 'http://admin:pass@localhost:5984/builds';
 //////////////////////////////// HACK
 
 const buildDb = new PouchDB(buildDbUrl);
@@ -41,10 +41,15 @@ module.exports = (buildInfo, username) => {
   return buildDb
     .get(ddocName(buildInfo), { attachments:true })
     .catch(err => {
+      console.log('GOT ERROR', err);
       if (err.status === 404) {
         err = new Error(`Version not found: ${buildInfo.version}`);
         err.expected = true;
       }
+      // FIXME: stop this from causing an unhandled promise rejection
+      //        It shouldn't do this because we catch again in server.js:194
+      //        Exceptions thrown in other parts (such as targetDb.put(newDDoc))
+      //        correctly flow to the second catch, but this one doesn't?
       throw err;
     })
     .then(newDdoc => {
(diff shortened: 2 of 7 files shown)


controllers/upgrade.js
```diff
@@ -0,0 +1,45 @@
...
+    .get(version, { attachments:true })
+    .then(newDdoc => console.log('upgrade()', 'Fetched newDdoc.') || newDdoc)
+
+    .then(newDdoc => console.log('upgrade()', 'Fetching oldDdoc…') || newDdoc)
+    .then(newDdoc =>
+      targetDb.get('_design/medic')
+        .then(oldDdoc => console.log('upgrade()', 'Fetched oldDdoc.') || oldDdoc)
+
+        .then(oldDdoc => {
+          newDdoc.app_settings = oldDdoc.app_settings;
```
SCdF: Interested in your thoughts @garethbowen, do you think this is the right logic? Is this the only thing we should be preserving? Are we preserving too much?

controllers/upgrade.js
```diff
@@ -0,0 +1,45 @@
...
+            version: version,
+          };
+
+          console.log('upgrade()', 'Uploading new ddoc…');
+          return targetDb.put(newDdoc)
+            .then(ret => console.log('upgrade()', 'newDdoc uploaded.') || ret);
+
+        }))
+    .catch(err => {
+      if (err.status === 404) {
```
garethbowen: Not sure why you're turning the 404 into a 500. 404 feels like the right status for "version not found". It'd be better not to catch this at all and remove the odd "expected server error" handling in server-utils.

server.js
```diff
@@ -173,6 +174,17 @@ app.get('/api/auth/:path', function(req, res) {
   });
 });
 
+app.post('/api/upgrade', jsonParser, (req, res) => {
+  auth.check(req, '_admin', null, (err, userCtx) => {
```
garethbowen: Expose `isDbAdmin` from auth for readability.

controllers/upgrade.js
```diff
@@ -0,0 +1,45 @@
...
+const buildDb = new PouchDB(buildDbUrl);
+const targetDbUrl = process.env.COUCH_URL;
+const targetDb = new PouchDB(targetDbUrl);
+
+module.exports = (version, username) => {
+  console.log('upgrade()', `Upgrading to ${version}…`);
+
+  console.log('upgrade()', 'Fetching newDdoc…');
+  return buildDb
+    .get(version, { attachments:true })
```
garethbowen: If you use just the version as the ID then you can only publish one product to each market. Is this on purpose? Marc and I have talked about publishing medic-reporter to horti too so we can get rid of dashboard from the instances. This would require a separate builds db. Can we instead use "[product-name]:[version]" as the key?

controllers/upgrade.js
```diff
@@ -0,0 +1,45 @@
...
+    .then(newDdoc => console.log('upgrade()', 'Fetched newDdoc.') || newDdoc)
+
+    .then(newDdoc => console.log('upgrade()', 'Fetching oldDdoc…') || newDdoc)
```
garethbowen: This log style is unnecessarily hard to read and will keep throwing the execution on the end of the execution queue. Just make the following `.then` block do the logging before calling `targetDb.get('_design/medic')`.

controllers/upgrade.js
```diff
@@ -0,0 +1,45 @@
...
+    .then(newDdoc =>
+      targetDb.get('_design/medic')
+        .then(oldDdoc => console.log('upgrade()', 'Fetched oldDdoc.') || oldDdoc)
```
garethbowen: As above.

SCdF:
I thought about writing tests, but all I can think of is a unit tests with heavy mocking confirming the single flow of code does what it does. Might be missing something.

controllers/upgrade.js
garethbowen: I think it's close, but if we ship new settings with defaults they'll be lost on upgrade. We need some sort of deep `_.defaults` so if we add a new nested configuration it'll merge in seamlessly. Something like (untested)...

```js
const deepDefaults = (oldObj, newObj) => {
  Object.keys(newObj).forEach(key => {
    if (typeof oldObj[key] === 'undefined') {
      return;
    }
    if (typeof oldObj[key] === 'object') {
      deepDefaults(oldObj[key], newObj[key]);
    } else {
      newObj[key] = oldObj[key];
    }
  });
  return newObj;
}
```

controllers/upgrade.js
```diff
@@ -0,0 +1,45 @@
...
+  console.log('upgrade()', `Upgrading to ${version}…`);
+
+  console.log('upgrade()', 'Fetching newDdoc…');
+  return buildDb
+    .get(version, { attachments:true })
+    .then(newDdoc => console.log('upgrade()', 'Fetched newDdoc.') || newDdoc)
+
+    .then(newDdoc => console.log('upgrade()', 'Fetching oldDdoc…') || newDdoc)
+    .then(newDdoc =>
+      targetDb.get('_design/medic')
```
garethbowen: Again I think we'll want to be able to upgrade other apps here so expose the `medic` bit in the API as the app name. Alternatively use the _id from `newDdoc`.

controllers/upgrade.js
```diff
@@ -0,0 +1,45 @@
...
+        .then(oldDdoc => console.log('upgrade()', 'Fetched oldDdoc.') || oldDdoc)
+
+        .then(oldDdoc => {
+          newDdoc.app_settings = oldDdoc.app_settings;
+          newDdoc._id = oldDdoc._id;
```
garethbowen: I think it'd be better to fail if this doesn't match.

garethbowen: I think it should fail if the IDs don't match.

controllers/upgrade.js
SCdF: Interesting! We hadn't thought of this. Yes we should definitely look into this. Let me think on it and I'll make a note to either solve it or raise a ticket before I'm done.

controllers/upgrade.js
SCdF: Done

controllers/upgrade.js
SCdF: I've changed the api medic-api exposes to expect the following JSON body:
```json
{
  "build": {
    "namespace": "medic",
    "application": "medic",
    "version": "1.2.3-beta.4"
  }
}
```

(as an example). We'll error if you don't pass that namespace and application, but it allows us to expand out later and pull this code out of api if we want to.

controllers/upgrade.js
SCdF: @garethbowen Defaults are dealt with in API already. See [config.js](https://github.com/medic/medic-api/blob/master/config.js#L59-L81). This runs initially and live on a listen.

If this logic isn't correct we should make it correct, or delete it, as it certainly gives the impression it fulfils the same function that your code above does.

garethbowen: You're right - the defaults in api should work.

controllers/upgrade.js
SCdF: done (I'm failing API if you don't pass `medic:medic` as your ns / app for now)

controllers/upgrade.js
SCdF: Did you have any thoughts on this @alxndrsn ?

server.js
SCdF: done

controllers/upgrade.js
alxndrsn: i'd have thought it'd be safe to assume that you're referring to `medic:medic:` in our app - surely the prefix is just to allow forwards-compatibility of the build db itself?

controllers/upgrade.js
alxndrsn: I'd say the code is wrong anyway - `Version not found: ${version}` should only be returned as the error message for the `buildsDb.get(version, { attachments:true })` call.  I'm not really convinced `404` is the correct response here, as the endpoint itself is available, but `400` would definitely be more appropriate the a 5xx error.

controllers/upgrade.js
SCdF: Yeah I think that's a fair point

controllers/upgrade.js
```diff
@@ -0,0 +1,76 @@
...
+    .catch(err => {
+      console.log('GOT ERROR', err);
+      if (err.status === 404) {
+        err = new Error(`Version not found: ${buildInfo.version}`);
+        err.expected = true;
+      }
+      // FIXME: why does this from causing an unhandled promise rejection?
+      //        It shouldn't do this because we catch again in server.js:194
+      //        Exceptions thrown in other parts (such as targetDb.put(newDDoc))
+      //        correctly flow to the second catch, but this one doesn't?
```
SCdF: NB: I'll investigate this next week, and fix / delete the comment / raise a ticket before merge

controllers/upgrade.js
```diff
@@ -0,0 +1,76 @@
...
+            user: username,
+            version: buildInfo.version,
+          };
+
+          console.log('upgrade()', `Staging new ddoc as ${newDdoc._id}`);
+          // TODO: if we have already pushed a staged ddoc this can cause a 409
+          //       Do we want to deal with this? Do we want to delete the existing
+          //       ddoc first? What if horti is in the middle of dealing with it?
+          //       Horti deletes this document once it's copied it into the right
+          //       place, so maybe leave this / make the error messag clearer
```
SCdF: Up for discussion.

Nuke what is already there:
 - Pros: Means a stuck staged build doesn't have to be manually deleted
 - Cons: No debouncing / other issues could screw things up (this may not be true, needs testing)

Throw error:
 - Pros: safer
 - Cons: pita if you're just the average user and you don't know what's going on

README.md
```diff
@@ -1237,3 +1237,32 @@ DELETE /api/v1/users/mary
...
+POST /api/v1/upgrade
+{
+  "build": {
+    "namespace": "medic",
+    "application": "medic",
+    "version": "3.0.0-beta.1"
+  }
+}
+```
+
```
mandric: I'm not crazy about this API. I'd prefer making it similar to a curl command or replication:
```
/api/v1/upgrade

{ 
  source: "https://https://staging.dev.medicmobile.org/_couch/builds/docID",
  target: "/_couch/medic/_design/medic"
}
```
source and target are both just URLs or paths.

controllers/upgrade.js
```diff
@@ -0,0 +1,76 @@
...
+
+      console.log('upgrade()', 'Fetching oldDdoc…');
+      return targetDb.get('_design/medic')
+        .then(oldDdoc => {
+          console.log('upgrade()', 'Fetched oldDdoc.');
+
+          newDdoc._id = stagedDdocName(oldDdoc._id);
+          delete newDdoc._rev;
+
+          newDdoc.deploy_info = {
```
mandric: I think I would rather this get recorded in the database as a meta doc rather then modifying the ddoc.

SCdF: I think we could do that. Something like `horticulturalist_deploy_info` (To scope it to this tool, just in case a more generically named document already exists).

On the plus side, it would mean we don't have to download the entire ddoc to read the value in the UI.
On the minus side, it means that you lose the binding between the ddoc (which is basically your application) and the meta-data about that. Which means bugs / edge cases could occur where they get out of sync (ie you deploy the ddoc some other way and horti inaccurately reports that you deployed it with horti, at some version, at some time).

What are you pros around moving it out of the ddoc?

mandric: The pros for me are organizational and separation of concerns/cleanliness.  I don't think deployment data has anything to do with the app itself.  The app (ddoc) should already have all the information we need on it when it gets built/released like commit hash and build time, and after that it never gets modified (app_settings is the only exception atm but is also a hack that probably needs to get fixed eventually).  All the meta data we need about the app/ddoc becomes part of the published asset and deployment data is managed separately.  This change seems to muddy the waters.  The ddoc to me feels like it should be treated like a binary, so modifying it when we install it seems weird, I'm hesitant.

README.md
mandric: I would agree with you if it didn't make the code more complex because  of handling and documenting namespace and app name in all the places.  Seems more logical to me to keep it more simple and flexible as a first cut and lock it down further later when these needs become more clear.  But as I look at it more it is growing on me.

controllers/upgrade.js
garethbowen: I think nuke it. Otherwise any bug in Horti will effectively brick the install process.

server-utils.js
```diff
@@ -86,11 +86,13 @@ module.exports = {
...
-   */
   serverError: function(err, req, res) {
     console.error('Server error:\n', err);
-    respond(req, res, 500, 'Server error');
+    if (err.expected) {
+      const status = err.status || 500;
+      respond(req, res, status, err.message);
+    } else {
+      respond(req, res, 500, 'Server error');
+    }
```
garethbowen: I really don't like this. The functions aren't well named, but the documentation that you deleted was there for a reason. Use the `error` function is it's maybe expected, maybe not. Use the `serverError` function if it's unexpected.

server.js
```diff
@@ -173,6 +174,27 @@ app.get('/api/auth/:path', function(req, res) {
...
+    if (err) {
+      return serverUtils.error(err, req, res);
+    }
+
+    var buildInfo = req.body.build;
+    if (!buildInfo) {
+      return serverUtils.error({
+        message: 'You must provide a build info body',
+        expected: true,
```
garethbowen: This isn't needed - `serverUtils.error` will detect that it's expected I think.

server-utils.js
SCdF: OK I'm going to revert any of the error stuff that's been done (FWIW they were in the original prototype), as that's the easiest way to get this 🚋  moving.

I think we want to look at cleaning up this stuff at some point though, it's unhelpful that the default failure in api is to throw a non-json string out. Unless I'm misunderstanding how to use it.

README.md
SCdF: > Seems more logical to me to keep it more simple and flexible as a first cut and lock it down further later when these needs become more clear

Heh, that is the direct opposite of what I think! As in, I think we should hard-code stuff so we're not building in flexibility we have to make decisions about until we need that flexibility. Specifically , like the last three bullet points in the comment above. We'd have to make decisions about that, consider security etc. I'd rather Just Not for now, at least until we've used it for a couple of months.

controllers/upgrade.js
SCdF: > I don't think deployment data has anything to do with the app itself.

Hmmmmmmmmm… I _sort of_ agree with this. On the other hand, I really don't like the edge cases that come out the deployment data *not* being attached to the app (as mentioned above).

Are you OK with us just going with this for now and then considering it later once we've used this for awhile? We're going to have to move stuff around here when we drop Kanso anyway.

README.md
alxndrsn: > I'd rather Just Not for now
+1 to this - let's get it in use, and build in features as they're needed

controllers/upgrade.js
mandric: Yeah it's fine, just my $0.02.

README.md
mandric: @alxndrsn I know the code has already been written but afaict, I'm asking to remove features, not add new ones, because it's always more work and unlikely to be removed later.

SCdF: OK I think I'm going to leave this as is for now.

I think that you're suggestion Milan is a possible correct implementations of this api.

But there are quite a few possible approaches we could take, and this is not a discussion we've had yet. I don't think there is value to moving to it without having these discussions, as there are definite downsides to it, which I half mentioned above[1]. I don't think it would help to simplify the code we already have either unfortunately, because we'd still need to hardcode our understanding of these various urls and version concepts in, to make sure people are only using it in the locked down way that we currently support. 

We can definitely change this later. If we stick with Horti and want to use it with other medic apps we *will* change it later. Now is just not the time.

[1] One obvious specific is the implication that you can upgrade a deployed app from any build server, because you're passing it on every upgrade. That seems really unsafe to me, unless we supported signing packages or something, because it removes ownership of our releases from us. I would favour configuring the build server in another location. There are probably more to consider, the overall point is that the consideration hasn't been done yet

PR #190, opened 2017-08-09
Default null translation values to key

garethbowen:
# Description

If the messages-xy.properties file has an empty translation value
it will be parsed and stored in our messages-xy doc as null. When
rendering using the angular directive this will safely fall back to
rendering the key, however when translating with the $translate
service an error is thrown. This change sets the value in the
database to the key so the $translate service behaves consistently
with the angular directive.

medic/medic-webapp#3753

# Review checklist


SCdF:
Tests passed (at time of writing, 6hrs after this PR was created the GH / travis integration still wasn't sure):
image

PR #192, opened 2017-08-15
Ensure all languages have all keys

garethbowen:
# Description

Uses English as the canonical set of translation keys and ensures
that any languages that are shipped with an incomplete set are
updated to contain all known keys.

medic/medic-webapp#3753

# Review checklist

### DIFF
--- a/translations.js
+++ b/translations.js
@@ -47,11 +47,22 @@ var createDoc = function(attachment) {
 
 var merge = function(attachments, backups, docs) {
   var updatedDocs = [];
+  var knownKeys = [];
+  var english = _.findWhere(attachments, { code: 'en' });
+  if (english) {
+    knownKeys = Object.keys(english.values);
+  }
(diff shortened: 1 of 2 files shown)


translations.js
```diff
@@ -47,11 +47,22 @@ var createDoc = function(attachment) {
 
 var merge = function(attachments, backups, docs) {
   var updatedDocs = [];
+  var knownKeys = [];
+  var english = _.findWhere(attachments, { code: 'en' });
+  if (english) {
```
SCdF: Surely we would never abandon The Queen!

SCdF: sacré bleu!!

translations.js
```diff
@@ -47,11 +47,22 @@ var createDoc = function(attachment) {
```
SCdF: This is a pretty small file, and we've been changing it quite a bit recently. Time to ES6-a-fy?

PR #193, opened 2017-08-16
Escape invalid characters when creating meta db

garethbowen:
# Description

The invalid characters are converted to their UTF code to reduce
the likelihood of collisions with other databases.

The create user api has been updated to fail validation if the
username has unusual characters.

medic/medic-webapp#3778

# Review checklist

### DIFF
--- a/controllers/users.js
+++ b/controllers/users.js
@@ -5,7 +5,8 @@ const async = require('async'),
       places = require('./places'),
       db = require('../db'),
       PASSWORD_MINIMUM_LENGTH = 8,
-      PASSWORD_MINIMUM_SCORE = 50;
+      PASSWORD_MINIMUM_SCORE = 50,
+      USERNAME_WHITELIST = /^[a-z0-9_-]+$/;
 
 const USER_EDITABLE_FIELDS = [
   'password',
@@ -128,12 +129,12 @@ var validateUserSettings = function(id, callback) {
 };
 
 var validateNewUsername = function(username, callback) {
+  if (!USERNAME_WHITELIST.test(username)) {
+    return callback(error400('Invalid user name. Valid characters are lower case letters, numbers, underscore (_), and hyphen (-).'));
+  }
   var id = createID(username);
   var error = function() {
-    return {
-      code: 400,
-      message: 'Username "' + username + '" already taken.'
-    };
+    return error400('Username "' + username + '" already taken.');
   };
   async.series([
     function(cb) {
--- a/lib/user-db.js
+++ b/lib/user-db.js
@@ -1,6 +1,10 @@
 const db = require('../db'),
-      async = require('async'),
-      DB_NAME_BLACKLIST = /[^a-z0-9_\$\(\)\+\-/]/g;
+      async = require('async');
+
+// Regex to test for characters that are invalid in db names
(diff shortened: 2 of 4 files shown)


controllers/users.js
```diff
@@ -5,7 +5,8 @@ const async = require('async'),
       places = require('./places'),
       db = require('../db'),
       PASSWORD_MINIMUM_LENGTH = 8,
-      PASSWORD_MINIMUM_SCORE = 50;
+      PASSWORD_MINIMUM_SCORE = 50,
+      USERNAME_WHITELIST = /^[a-z0-9_-]+$/;
```
alxndrsn: Do we want to impose a min length of more than one character?

lib/user-db.js
```diff
@@ -1,5 +1,6 @@
 const db = require('../db'),
-      async = require('async');
+      async = require('async'),
+      DB_NAME_BLACKLIST = /[^a-z0-9_\$\(\)\+\-/]/g;
```
alxndrsn: I'm terrible with regexes but I'm struggling reading this one.  I don't think special characters need escaping inside square brackets.

Also, where does this blacklist come from?  Is this a couch restriction, or something we've decided on ourselves?  Might be nice to document that somehow, e.g. a link to relevant couch docs from this code.

tests/unit/lib/user-db.js
```diff
@@ -7,6 +7,26 @@ exports.tearDown = callback => {
...
+  const actual = lib.getDbName(given);
+  test.equals(actual, expected);
+  test.done();
+};
+
+exports['getDbName escapes invalid characters - #3778'] = test => {
+  // Only lowercase characters (a-z), digits (0-9), and any of the characters _, $, (, ), +, -, and / are allowed.
+  const valid   = 'abc123_$()+-/';
+  const invalid = '.<>^,?!';
+  const escaped = '(46)(60)(62)(94)(44)(63)(33)';
```
alxndrsn: Is this really how we want to escape these characters?  Would something more standard, e.g. standard URL-encoding (percent-sign + hex) be usable?

controllers/users.js
garethbowen: Can you elaborate on why that would be useful?

lib/user-db.js
garethbowen: I'll play around with cleaning it up, and also link to the documentation.

tests/unit/lib/user-db.js
garethbowen: `%` is not a valid character either so that wouldn't help. The only punctuation we can use is: `_`, `$`, `(`, `)`, `+`, `-`, or `/`. I thought parens were good but open to suggestion here. I could also switch from decimal to hex if you think that's better?

lib/user-db.js
garethbowen: It was only the hyphen that needed escaping because it interpreted `+-/` as being a range. I've moved the hyphen to the end of the group and it just works with no escaping. The unit test fails if this gets messed up so we should be safe from regressions.

controllers/users.js
alxndrsn: No, I guess it doesn't really matter :slightly_smiling_face:

tests/unit/lib/user-db.js
alxndrsn: I think hex might be more obvious, but I don't think the issue's important enough to spend too much time worrying.

PR #195, opened 2017-08-25
Gracefully handle errors from request lib

garethbowen:
# Description

This improves our error handling so errors being thrown from the
request library do not crash API. From now on any 5xx error is to
be treated as a server error.

medic/medic-webapp#3821

# Review checklist

### DIFF
--- a/tests/unit/server-utils.js
+++ b/tests/unit/server-utils.js
@@ -56,10 +56,10 @@ exports['error handles node errors'] = test => {
   const end = sinon.stub(res, 'end');
   serverUtils.error({ code: 503, message: 'some error' }, req, res);
   test.equals(writeHead.callCount, 1);
-  test.equals(writeHead.args[0][0], 503);
+  test.equals(writeHead.args[0][0], 500);
   test.equals(writeHead.args[0][1]['Content-Type'], 'text/plain');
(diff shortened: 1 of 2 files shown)


tests/unit/server-utils.js
```diff
@@ -69,10 +69,35 @@ exports['error handles nano errors'] = test => {
...
-  test.equals(writeHead.args[0][0], 503);
+  test.equals(writeHead.args[0][0], 500);
   test.equals(writeHead.args[0][1]['Content-Type'], 'text/plain');
   test.equals(end.callCount, 1);
-  test.equals(end.args[0][0], 'some error');
+  test.equals(end.args[0][0], 'Server error');
+  test.done();
+};
+
+exports['error handles request errors - #3821'] = test => {
```
alxndrsn: I don't understand the title of this test.  I thought the change was to handle upstream server errors, not request errors.  

Also, given the repeat of the word "error" in this test name, it might be clearer to suffix the function name with `()`.

garethbowen: Yes it should handle all errors but the example I've used in the test is a copy of the one you get from the `request` library. I'll change the title and add a comment explaining where the example comes from.

PR #198, opened 2017-08-25
Propogate unknown error messages to client

alxndrsn:
Closes medic/medic-webapp#2994

### DIFF
--- a/tests/unit/server-utils.js
+++ b/tests/unit/server-utils.js
@@ -85,7 +85,7 @@ exports['error handles unknown errors'] = test => {
   test.equals(writeHead.args[0][0], 500);
   test.equals(writeHead.args[0][1]['Content-Type'], 'text/plain');
   test.equals(end.callCount, 1);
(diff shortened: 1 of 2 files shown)


tests/unit/server-utils.js
```diff
@@ -152,7 +152,7 @@ exports['serverError does not leak errors information to the client'] = test =>
   test.equals(writeHead.args[0][0], 500);
   test.equals(writeHead.args[0][1]['Content-Type'], 'text/plain');
   test.equals(end.callCount, 1);
-  test.equals(end.args[0][0], 'Server error');
+  test.equals(end.args[0][0], 'Server error: boom');
```
garethbowen: The title of this test now doesn't match with this assertion. The test is there to check for the [security risk of leaking information](https://www.owasp.org/index.php/Error_Handling#Generic_error_messages). My concern is because we don't know for sure what the error message is we could be exposing sensitive information, worst case basic auth credentials in environment variable URLs.

PR #199, opened 2017-08-29
Add errors which are explicitly displayable to client

alxndrsn:
Add errors which are explicitly displayable to client

Issue medic/medic-webapp#2994


alxndrsn:
Is this what you meant?  If so, I'm not a fan - more code and harder to read.  Although an advantage is that it won't have an extra entry in the stacktrace for `./public-error.js`.

alxndrsn:
Merged manually.

PR #203, opened 2017-10-05
Creates a migration to install the default icons

garethbowen:
# Description

Modifies the 'resources' doc with the default icons for clinic,
health_center, district_hospital, and person so we're not limited
to the font-awesome icons.

medic/medic-webapp#3945

# Review checklist


garethbowen:
> Looks fine if the resources doc is guaranteed to exist :¬)

There's a migration that adds it. If it's been deleted other bad things will happen...

> Is there a reason for rolling your own attachment encoding instead of using db.putAttachment()?

I didn't want to make 4 requests and 4 revs when I could make one.

> Are the other attachments loaded fully when the resources doc is fetched? If not, is it OK to add new base64 attachments?

Interesting... it worked in my testing, so I guess it's ok. Hopefully any issues will be picked up in AT or RT.

> Does it matter that existing attachments with the same name would be overwritten?

Yeah, it probably does. I'll modify it so it doesn't change existing resources. That would be unexpected. Good catch.

PR #210, opened 2018-04-23
Add check for missing changes results

alxndrsn:
Backport of fix for medic/medic-webapp#3099 to close medic/medic-webapp#4460.


alxndrsn:
(This fix was already included in the `2.15.x` branch.)

alxndrsn:
@SCdF is it worth backporting to 2.13 as well?

SCdF:
@alxndrsn I think if it comes up we can, otherwise I'm not sure it's too important

